PR #9, opened 2019-05-30
Basic auth for cLoki

lansio:
Same as basic auth in Loki


lmangani:
Thanks @lansio much needed and appreciated!

lmangani:
@lansio I set the Fastify auth module loading conditional to the ENV being set. I see you hardcoded authentication details in the PR. Are those defaults in `loki` now or shall we set to unathenticated unless the preference is expressed?

lansio:
Hmm, I guess your option is preferable.
Hardcode was my fault, of course, auth should be configurable.

many thanks for fix!

PR #15, opened 2020-02-19
Update to latest Loki API

lmangani:


lmangani:
Anyone out there willing to give this a try and provide some feedback?

lmangani:
Anyone out there willing to give this a try and provide some feedback?

Slach:
thanks a lot for your efforts!

lmangani:
@Slach thanks to you for the support! if you get a chance to test it, let us know so we can iron out any problem. Basic insert via API and Grafana queries seem to work fine. 

NOTE: Still implementing other parts of the new API and protobuf support is unfortunately still broken - we need a special genius to convert the group definitions in a way nodejs could leverage.

PR #25, opened 2021-02-05
loki push api fix for v1

shawel:


lmangani:
Thanks for the patch + format @shawel (could you tell I use a console to edit code?) ;)

lmangani (approved):

shawel:
Hah! Not really I have a setup with sublime that auto-formats.

PR #49, opened 2021-11-03
Watcher

akvlad:
https://github.com/lmangani/cLoki/issues/30
LIVE mode support for cLoki


Dletta (approved):
Good to go from testing side as well

PR #50, opened 2021-11-05
Minor features optimization

akvlad:
minor features:
- series request
- GET ready request

### DIFF
--- a/cloki.js
+++ b/cloki.js
@@ -144,6 +144,8 @@ fastify.get("/loki/api/v1/series", handler_series);
 
 fastify.get('/loki/api/v1/tail', {websocket:true}, require('./lib/handlers/tail').bind(this));
 
+fastify.get('/ready', require('./lib/handlers/ready').bind(this));
+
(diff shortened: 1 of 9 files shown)


cloki.js
```diff
@@ -109,8 +109,9 @@ const handler_404 = require('./lib/handlers/404.js').bind(this);
 fastify.setNotFoundHandler(handler_404);
 
 /* Hello cloki test API */
-const handler_hello = require('./lib/handlers/hello.js').bind(this);
+const handler_hello = require('./lib/handlers/ready').bind(this);
 fastify.get("/hello", handler_hello);
+fastify.get("/ready", handler_hello);
```
lmangani, line 114: Some clients use the /hello endpoint, we should decommission it at a later time. 
Could we keep /hello as an alias for the /ready handler?

akvlad: Exactly the thing I did.
Or you want this request to response "cloki"?

Dletta (approved):

PR #53, opened 2021-11-08
Standard Style Linting

Dletta:


Dletta (commented):
Reviewed, should be good

PR #57, opened 2021-11-10
CliQL improved matching

lmangani:
Improved matching for CliQL (native clickhouse) query parsing in case of where conditions


Dletta (approved):
I see no issues with the changes

PR #58, opened 2021-11-11
labels to time series extractor plugin

akvlad:
parser plugins simple implementation.
label_to_row function according to the request


lmangani:
@Dletta please document in context of merging

akvlad:
@lmangani  I'll write everything in our Wiki after we accept the PR.

Dletta (approved):

PR #63, opened 2021-11-20
WIP: switch to snappy

Dletta:
Needs testing with a protobuf pipe


akvlad:
Tested against centos7 && ubuntu 20.04.3
OK

PR #65, opened 2021-11-21
Refresh /query API

lmangani:
The current /query API needs to be refreshed with the new method and transpiler.

API Specs: https://grafana.com/docs/loki/latest/api/#get-lokiapiv1query

Review/Patching required @akvlad


lmangani:
@akvlad @Dletta is this ready to be merged?

Dletta (approved):
tested working

PR #72, opened 2021-11-29
fix parsing of DEBUG env var

mr-karan:
Closes https://github.com/lmangani/cLoki/issues/71

P.S. I don't know the proper JS coding style, if there's a better way to fix this, please feel free!

### DIFF
--- a/cloki.js
+++ b/cloki.js
@@ -5,8 +5,10 @@
  * (C) 2018-2021 QXIP BV
  */
 
-this.debug = process.env.DEBUG || false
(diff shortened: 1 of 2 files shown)


lmangani:
This does not appear to be an issue and the existing code works as expected. Opening an issue is suggested before a PR.
Note: set this variable to anything (not necessarily a bool) for it to be triggered in general

mr-karan:
I've set `DEBUG=false` and it still logs the debug lines.

PR #76, opened 2021-11-30
Misc Cosmetic Fixes

lmangani:
- Mixed PR with cosmetic fixes to readme, examples
- Change response of PUSH API to 204 (no content)


mr-karan:
@lmangani Just noticed that you changed `res.send(200)` to `res.send(204)`. That won't work if you are intending to fix https://github.com/lmangani/cLoki/issues/74, because the default status code is HTTP 200. So, we need to use `res.code(204)` and `.send()` should ideally be empty because HTTP 204 Indicates no content in response body.

>  The HTTP 204 No Content success status response code indicates that a request has succeeded, but that the client doesn't need to navigate away from its current page. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

lmangani:
@mr-karan feel free to try this PR which includes the response fix you suggested/requested

mr-karan:
@lmangani I'm running a patched version of `cLoki` with that fix already and it works :) Thanks

PR #79, opened 2021-12-02
Metrics support

akvlad:
`rate({test_id="1"} | unwrap_value [1s])`
unwrap_value statement offers a user to work with 'value' metrics field like with an unwrapped value.


Dletta (approved):
Testing, will merge after completion

PR #80, opened 2021-12-07
Logfmt

Dletta:
Log QL functionality for Logfmt, that parses the complete log line into an object (reminiscient of JSON parse)


akvlad:
Approved

PR #86, opened 2021-12-24
wasm fail fixes; undefined entry fail fix

akvlad:
some  fail bugfixes


Dletta (approved):
Tested, working  as expected with flag ```LINE_FMT=go_native```

PR #93, opened 2022-01-12
legacy matching

lmangani:


lmangani:
@Dletta CI is failing where it shouldn't please review the Tempo e2e test

lmangani:
@Dletta The trace span is 24h and the timestamp of the insert is out of range. Needs a dynamic timestamp.

PR #97, opened 2022-01-24
multitenancy support init

akvlad:
multitenancy support


akvlad:
# Application changes
## Multitenancy Initialization
The initial env variable for the multitenancy support is 
`MULTITENANCY=ON`. 
It should be exported all the time when cLoki is used with organization IDs.
## Database structure for the multitenancy
The new term added to the cLoki infrastructure is *"home tenant"* The *home tenant* is the database with the `settings` table where all the tenants info is stored. When a new request is come to cLoki, the organization id is found in that `settings` table.
The home tenant is configured by the default cLoki env variables: CLICKHOUSE_SERVER, CLICKHOUSE_DB, etc.
If a request to cLoki doesn't have the X-Scope-OrgID header, then the request is considered as a request to the home tenant.
## Adding a new tenant
In order to add a new tenant you should:
- export all the default environment for the home tenant:
```
export CLICKHOUSE_DB=cloki
export MULTITENANCY=ON
```
- run cloki CLI with `add-tenant` command:
`node cloki.js add-tenant --org org1 --url http://localhost:8123/ --db db1 --rotate-samples-days 7 --rotate-time-series-days 7 --storage-policy spol1`

Arguments for the command:

| Arg | Required | Description |
| --- | --- | --- |
| --org | Yes | Org ID |
| --db | [Yes | Database name |
| --rotate-samples-days| No | Samples table TTL in days (default is 7)
| --rotate-time-series-days | No | Time series table TTL in days (default is 7)
| --storage-policy | No | Storage policy to apply for the tables (no storage policy by default)

## Some additional commands for the cloki CLI

`node cloki.js update-only` update all the databases to the latest version of the schema and  update the rotation policies.
`node cloki.js rotate-only` update all the databases to the latest version of the schema and  update the rotation policies.

lmangani:
Great start! The command method seems potentially a limiting factor.  A couple points for discussion:
- Are there secured API or query equivalents?
- How are they stored at the DB layer?

akvlad:
# Some new information about the cLoki infrastructure to be considered for the further development
Development of multitenancy for cLoki led to a vast code review and re-engineering, The main infrastructure changes are the next.

## SQL schema upgrade
The upgrade scripts are placed into a separate file https://github.com/lmangani/cLoki/blob/multitenancy/lib/db/maintain/scripts.sql ,
- The script is applied command by command from the top to the bottom. After each command is done its index is stored to the `ver` table for the tenant. So a command will not be applied twice.
- Each command is separated by semicolon and newline `;\n`. Using semicolons in the end of line for another cases is prohibited.
- Update commands already applied are prohibited to alter. New updates should be done via new "patch" commands. In other words, **NO ALTERATIONS AFTER GIT PUSH! APPEND ONLY**. 
- Inside the script you can use environment variables via Handlebars processor (https://github.com/lmangani/cLoki/blob/multitenancy/lib/db/maintain/scripts.sql#L10) as well as another handlebars functions.

## Clickhouse client
Clickhouse client is redesigned to be an object instead of the module it was before. 
The client class and its subclasses are implemented here:
- https://github.com/lmangani/cLoki/blob/multitenancy/lib/db/clickhouse.js#L238 - main client class (single-tenant)
- https://github.com/lmangani/cLoki/blob/multitenancy/lib/multitenancy/index.js#L49 - client class for multitenancy support
- https://github.com/lmangani/cLoki/blob/multitenancy/lib/db/clickhouse_alerting.js#L11 - subclass for alerting

The proper clickhouse client should be received in one of two ways:
- using `await request.client()` call inside a fastify http handler. https://github.com/lmangani/cLoki/blob/multitenancy/lib/handlers/label.js#L16 It returns an appropriate clickhouse client object for the tenant specified in the request (if multitenancy is ON) or a default client (if not).
- using `multitenancy.getClient(orgID)` method directly https://github.com/lmangani/cLoki/blob/multitenancy/lib/multitenancy/index.js#L11

Issue #68, opened 2021-11-29
Provide versioned docker images


mr-karan:
Hi

Firstly, thanks for building and open sourcing this project! It's super amazing and I came here after watching [this talk](https://www.youtube.com/watch?v=E1yuPXNz5Fo). Clickhouse fan here as well and would love to try this project in my org.

I am looking to set this up in production and aiming at a `docker` setup. I think it'll be really nice if the [docker images](https://hub.docker.com/r/qxip/cloki/tags) are published with proper tags and not `:latest`. Since this project is in beta stage, it's natural for upgrades to maybe break an existing version so just to keep it safe, a tagged version would be much appreciated.

Thanks!

lmangani:
Thanks @mr-karan - good point indeed.  Since we are currently pushing features/fixes intensively and we pretty much need all users on the latest version and support validation, we're intentionally sticking with `latest` tag but once the release cycle will switch to stable feature releases we'll modify the GitHub action to push tagged containers.

Let's keep this issue open as a gateway reminder towards this step.

Thanks!

lmangani:
_TODO:_ change [docker tag](https://github.com/lmangani/cLoki/blob/master/.github/workflows/bump_version.yml#L86) to `${{ steps.version-bump.outputs.newTag }}`

derlaft:
@lmangani there's already "v2" released with some breaking changes - is it finally time to enable docker tags?

lmangani:
@derlaft yes and good news, we've been testing this on a separate staging and it'll soon be merged into the actions

PR #103, opened 2022-02-04
enable docker tags on builds

danjenkins:
in this case, we'll have `latest` and `sha1` hashes as we don't run the action unless we're on master branch... but... I left the action code in there so that when you do change it to be able to run off more branches/tags/releases it'll auto work


Dletta:
Referencing #68

PR #108, opened 2022-02-07
Move to a proper logging library

danjenkins:
This change is not well tested yet.

This change brings in a proper log library called `pino`

It's already a part of `fastify`

This enables us to be able to log things out in a much nicer way for, ironically, log ingestion among other things
This change also adds a new env var called `LOG_LEVEL` so that people running the software have greater control over logging
For backwards compatibility `DEBUG` env var will set `LOG_LEVEL` to debug

Most obvious change will be that fastify logging is no longer disabled and will do an info level log on every request input and output
this can be disabled using the flag `disableRequestLogging` in fastify server setup

If you want to be able to see logs that aren't designed to be machine readable all of the time you can start up the process with `npm run pretty` which will start up the process with `pino-pretty` analysing the logs

Oh and I changed `CLOKI_LOGIN` and `CLOKI_PASSWORD` to be `undefined` by default, not `false` - it's bad convention to change a type of variable - you were going from `boolean` to be `string`, `undefined` did the same job.

OH and you'll see i changed a lot of `catch(e)` to `catch(err)` - that's because pino has a built in serializer for the key `err` and it should make it so all errors have keys of `err` instead of a mix of `e` and `err` and `error`

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Check out the [Wiki](https://github.com/lmangani/cLoki/wiki) for detailed instru
 Clone this repository, install with `npm`and run using `nodejs` 14.x *(or higher)*
 ```bash
 npm install
-CLICKHOUSE_SERVER="my.clickhouse.server" CLICKHOUSE_AUTH="default:password" CLICKHOUSE_DB="cloki" node ./cloki.js
+CLICKHOUSE_SERVER="my.clickhouse.server" CLICKHOUSE_AUTH="default:password" CLICKHOUSE_DB="cloki" node cloki.js
 ```
 ##### :busstop: NPM
 Install `cloki` as global package on your system using `npm`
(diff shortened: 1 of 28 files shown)


adubovikov:
@danjenkins Thank you so much! We are really very happy and appreciated for your contribution!

akvlad:
@danjenkins I have one question about this:
```
    } catch (err) {
      logger.error({ err }, 'AXIOS ERROR')
      this.rejects.forEach(r => r(err))
    }
```
Axios error objects are giant in text representation. They are several screens long. Does your commit have some output shortening mechanism instead of:
```
    } catch (e) {
      if (e.response) {
        console.log('AXIOS ERROR')
        console.log(e.message)
        console.log(e.response.status)
        console.log(e.response.data)
      } else {
```
?

akvlad:
Ok. It will decrease size of exceptions. Thank you @danjenkins .

PR #114, opened 2022-02-09
Replace short-hash with xxHash64 (should fix #113)

### LINKED ISSUE #113: Fingerprinting hashing algorimth is prone to collisions
After ingesting logs from the k8s cluster for quite some time I've noticed some strange results. When I was querying for one specific label, I was getting results which should clearly belong to a completely different one.

Some digging showed that the problem is collision:

```:) select date, fingerprint, hex(MD5(labels)) from time_series where fingerprint=523712935;

┌───────date─┬─fingerprint─┬─hex(MD5(labels))─────────────────┐
│ 2022-02-09 │   523712935 │ A2494F27CF46F29BCD88809CE8267F23 │
│ 2022-02-04 │   523712935 │ 57F427CE06BD67AC5E45BF62F94A5177 │
└────────────┴─────────────┴──────────────────────────────────┘
```

This is the current fingerprinting code:

https://github.com/lmangani/cLoki/blob/c40d1d9664523e60964a8d0b9338575c0a6d4720/lib/utils.js#L21-L26

It goes all the way here:

https://github.com/MatthewBarker/hash-string/blob/master/source/hash-string.js#L6-L16

It looks to me it's very suboptimal, since very fast, efficient, and much less prone for collisions hashes do exist. For example, xxHash64, also available in js: https://www.npmjs.com/package/xxhashjs.

derlaft:
https://github.com/lmangani/cLoki/issues/113

Changes:

* Replace `short-hash` library with `xxhashjs` library. Should result in superior collision resistance. 
* There might be some other options, I have not evaluated them. Selected this library because I know it works well and because js port exists
* There is also this port: https://www.npmjs.com/package/xxhash-addon/v/1.4.0
    - it looks like it might be faster, but it requires compiling some native stuff. wdyt?
* No idea what happens during migration. Looks to me it might just work (while maybe generating a number of other collisions). wdyt?

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -19,10 +19,11 @@ const labelParser = function (labels) {
   return { labels: output, regex: regex }
 }
 /* Fingerprinting */
-const shortHash = require('short-hash')
+const xxh = require('xxhashjs')
+const xxhSeed = 'cloki'
 const fingerPrint = function (text, hex) {
-  if (hex) return shortHash(text)
-  else return parseInt(shortHash(text), 16)
+  if (hex) return xxh.h64(text, xxhSeed).toString(16)
+  else return Number(xxh.h64(text, xxhSeed))
 }
 
 const toJSON = require('jsonic')
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,11 +33,11 @@
         "protocol-buffers-encodings": "^1.1.1",
         "record-cache": "^1.1.1",
         "scramjet": "^4.36.1",
-        "short-hash": "^1.0.0",
         "snappyjs": "^0.6.1",
         "stream-chain": "^2.2.4",
         "stream-json": "^1.7.3",
         "ws": "^8.4.0",
+        "xxhashjs": "^0.2.2",
         "yaml": "^1.10.2"
       },
       "bin": {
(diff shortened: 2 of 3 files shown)


derlaft:
>  2 workflows awaiting approval 
-_-

derlaft:
> Some tests are failing locally (even without the change)

Was missing `npm postinstall`, probably should be documented somewhere

lmangani:
@derlaft thanks for opening this issue. The problem is well known and the intention in general would to let users configure a hashing function of preference as long as it's natively supported by clickhouse for compatibility with other pipelines we need to support. Could you confirm if `xxhashjs` produces the same hashes as the clickhouse native xxHash64() function?

derlaft:
> Could you confirm if xxhashjs produces the same hashes as the clickhouse native xxHash64() function?

Looks like no

> with other pipelines we need to support

With which pipelines? Is it something custom?

lmangani:
@derlaft part of our design supports ingestion through other means (kafka, etc) and we MUST retain compatibility with the hashing function. I think we should simply support multiple hashing mechanism and let the client configuration decide which one to use when writing. The reading stage should have no challenges dealing with mixed-hash scenarios.

derlaft:
> and we MUST retain some level of compatibility with the native hashing function

Hashing compatibility does make sense. However, there's one more problem with that: currently JS implementation writes `labels` objects with keys sorted in random order.

> I think we should simply support multiple hashing mechanism and let the client configuration decide which one to use when writing

Should the current broken one be included in the list? That's the part I don't really understand "why" given it produces so many collisions on small amount of data.

derlaft:
Let's move discussion to https://github.com/lmangani/cLoki/issues/113

lmangani:
@derlaft to answer your question, the proposal is simply to decommission it gradually and to establish multiple hashing options for the future.

PR #115, opened 2022-02-09
XXH hash function

akvlad:
env configurable HASH function
`export HASH=XXH` if you want to use xxH64 function.


lmangani:
Should we default to XXH? 

Note to self: update README and Wiki

akvlad:
@lmangani I'd prefer to leave default `short-hash` untouched.

lib/utils.js
```diff
@@ -20,7 +20,15 @@ const labelParser = function (labels) {
...
 /* Fingerprinting */
 const shortHash = require('short-hash')
-const fingerPrint = function (text, hex) {
+const xxh = require('xxhashjs')
+const xxhSeed = 'cloki'
+const fingerPrint = function (text, hex, alg) {
+  alg = alg || process.env.HASH || 'short-hash'
+  switch (alg) {
+    case 'XXH':
+      if (hex) return xxh.h64(text, xxhSeed).toString(16)
```
derlaft, line 29: To keep compatibility with clickhouse seed is not needed

```suggestion
const fingerPrint = function (text, hex, alg) {
  alg = alg || process.env.HASH || 'short-hash'
  switch (alg) {
    case 'XXH':
      if (hex) return xxh.h64(text, '').toString(16)
```

lib/utils.js
```diff
@@ -20,7 +20,15 @@ const labelParser = function (labels) {
...
+      else return Number(xxh.h64(text, xxhSeed))
```
derlaft: `Number(UInt64)` seems to lose data:
```
> xxh.h64(0).update('test').digest().toString(10)
'5754696928334414137'
> Number(xxh.h64(0).update('test').digest())
5754696928334414000
```

Not sure (I am bad at javascript) - maybe a string can be returned here?

derlaft: Tests do pass with this (and forced XXH by default):

```suggestion
      else return xxh.h64(text, '').toString(10)
```

probably because inserting numbers as strings works for numeric columns in clickhouse. this way all 64 bits can be used.

README.md
```diff
@@ -180,6 +180,7 @@ The following ENV Variables can be used to control cLoki parameters and backend
 | TEMPO_TAGTRACE | false | Optional tagging of TraceID (expensive) |
 | DEBUG  			| false  	    | Debug Mode (for backwards compatibility) 		|
 | LOG_LEVEL  			| info  	    | Log Level  		|
+| HASH | short-hash | Hash function using for fingerprints. Currently supported `short-hash` and `XXH` (xxH64 function)
```
derlaft, line 183:
```suggestion
| HASH | short-hash | Hash function using for fingerprints. Currently supported `short-hash` and `XXH` (xxHash64 function). Default `short-hash` function is known to produce a significant number of collisions.
```

lib/utils.js
```diff
@@ -20,7 +20,14 @@ const labelParser = function (labels) {
 }
 /* Fingerprinting */
 const shortHash = require('short-hash')
-const fingerPrint = function (text, hex) {
+const xxh = require('xxhashjs')
+const fingerPrint = function (text, hex, alg) {
+  alg = alg || process.env.HASH || 'short-hash'
+  switch (alg) {
+    case 'XXH':
```
derlaft, line 27: naming is not consistent:

* `short-hash`
* `XXH`

For example, `XXH` could be renamed to `xxhash64`

derlaft:
https://www.npmjs.com/package/xxhash-wasm

hmm, there's also this lib

akvlad:
Well, tests seem passing for both hashes. @Dletta @derlaft please recheck carefully because alterations turned out to be wider than I expected.

lib/utils.js
```diff
@@ -20,7 +20,14 @@ const labelParser = function (labels) {
...
 /* Fingerprinting */
 const shortHash = require('short-hash')
-const fingerPrint = function (text, hex) {
+const xxh = require('xxhashjs')
+const fingerPrint = function (text, hex, alg) {
+  alg = alg || process.env.HASH || 'short-hash'
+  switch (alg) {
+    case 'xxhash64':
+      if (hex) return xxh.h64(text, '').toString(16)
+      else return BigInt(xxh.h64(text, '').toString())
```
derlaft, line 29: Looks like UINT64 -> String -> BigInt conversion can be replaced with UINT64 -> BigInt:

```suggestion
      if (hex) return xxh.h64(text, '').toString(16)
      else return BigInt(xxh.h64(text, '').digest())
```

derlaft: *fixed

lmangani: Did you see any measurable advantage in terms of performance?

derlaft: not sure if it's measurable, but this extra step is simply not needed (and can slow down things or be about as good in the best case)

lib/utils.js
```diff
@@ -20,7 +20,14 @@ const labelParser = function (labels) {
...
+  }
   if (hex) return shortHash(text)
   else return parseInt(shortHash(text), 16)
```
derlaft, line 32: as I mentioned in the issue this also can be improved. [Currently it expands to](https://github.com/joakimbeng/short-hash/blob/b7e1178b45b06bbf1697ba428b7da53664e96979/src/index.js#L5):
```
parseInt(hash(text).toString(16), 16)
```

There's no reason to do that, [hash-string](https://www.npmjs.com/package/hash-string) library can be directly used

lmangani:
@derlaft any chance to test the current PR and confirm if it creates any issues on your setup?

derlaft:
> @derlaft any chance to test the current PR and confirm if it creates any issues on your setup?

I need to build a docker image somehow... Let me try

derlaft:
```
Feb 09 13:44:27.075  WARN sink{component_kind="sink" component_name=loki component_type=loki}:request{request_id=101695}: vector::sinks::util::retries: Request timed out.
```

not working... too slow now

lmangani:
> 95}: vector::sinks::util::retries: Request timed out.

Weird and unexpected. Any errors on the cloki side or is it just backpressure of the hashing?

derlaft:
Maybe it was indeed a back pressure, works now:

```
{"level":30,"time":1644416901397,"pid":18,"hostname":"cloki-staging-deployment-c83bd120-7467c88496-sjw2j","name":"cloki","reqId":"req-am","res":{"statusCode":204},"responseTime":139.88854598999023,"msg":"request completed"}
3
{"level":30,"time":1644416901451,"pid":18,"hostname":"cloki-staging-deployment-c83bd120-7467c88496-sjw2j","name":"cloki","reqId":"req-an","res":{"statusCode":204},"responseTime":130.30693244934082,"msg":"request completed"}
2
{"level":30,"time":1644416901452,"pid":18,"hostname":"cloki-staging-deployment-c83bd120-7467c88496-sjw2j","name":"cloki","reqId":"req-ao","res":{"statusCode":204},"responseTime":98.58740234375,"msg":"request completed"}
```
Is `responseTime` seconds?

>  I see 3x performance reduce for our new hash. 

Don't see any difference so far, avg time is 150-500 (assuming seconds) here before and after the change (a bit strange).

akvlad:
@derlaft 
>Don't see any difference so far
I created a benchmark for 500+ fingerprints. It showed 3x performance reduce. 
Great to hear that it worked for you.

derlaft:
> I created a benchmark for 500+ fingerprints. It showed 3x performance reduce.

What about wasm version, how bad is it?

> It showed 3x performance reduce.

I am not sure, maybe I am looking at a wrong thing. How do you measure performance?

derlaft:
> wasm implementation helped a bit. Please recheck.

So far it's working, `responseTime` in logs do show similar average values as before

lib/db/throttler.js
```diff
@@ -16,6 +16,7 @@ class TimeoutThrottler {
       await this._flush()
       this.resolvers.forEach(r => r())
     } catch (err) {
+      console.log(err)
```
Dletta, line 19: @akvlad Needed still? Straggler from testing?

Dletta (approved):
Change has been tested against inserts and Grafana Dashboard (queries of all ranges LogQL)

lib/db/throttler.js
lmangani: lets clean it up or move to the new logger error log

PR #116, opened 2022-02-09
Docker + PM2 Clustering

lmangani:
#### Docker
- [x] pm2 ecosystem config in cluster mode with no ENV variables (shell or compose provided)
- [x] API requests distributed over multiple stateless threads based on available CPUs


lmangani:
@danjenkins would you have a chance to try this out and confirm if it works w/o requiring any config changes?

lmangani:
@derlaft would you have a chance to try this container build out and confirm if it works w/o requiring any config changes?

Dockerfile
```diff
@@ -1,15 +1,18 @@
...
-ENV BUILD 703030
+# Build ENV
+ENV BUILD 703033
 ENV PORT 3100
 
+# Build App
 COPY . /app
 WORKDIR /app
 RUN npm install
+RUN npm install pm2 -g
```
derlaft, line 12: If you want to improve caching and layer difference size between builds when dependencies are not changed you can do something like this:

```suggestion
# Build App
RUN mkdir /app
WORKDIR /app
COPY package.json /app
COPY package-lock.json /app
RUN npm install
RUN npm install pm2 -g
COPY . /app
```

lmangani: Good point. Indeed we intentionally avoided container optimizations so far (see the top ENV forcing rebuilds) due to the high number of changes and library changes in early stages but the time is right. For now I was primarily interested in making sure the switch from directly execution of cloki towards the pm2 execution didn't impact your ability to use the current ENV variables/compose/k8s you are using.

derlaft: > For now I was primarily interested in making sure the switch from directly execution of cloki towards the pm2 execution didn't impact your ability to use the current ENV variables/compose/k8s you are using.

Sure, will check that a bit later

akvlad: @lmangani @derlaft 
`RUN npm install pm2 -g`
should be before 
`COPY package.json /app`
to avoid reinstallation of pm2 each time when packaje.json is changed.
But as far as I know, these dockers are built quite rare and mostly with `--no-cache`, so...

derlaft: >  But as far as I know, these dockers are built quite rare and mostly with --no-cache, so...

1) it's useful for development/testing (I have to personally push test cLoki images to registry via slow DSL)

2) You are using docker build action, pushing directly to registry. This way cache can be exported together with layers ([BuildKit](https://github.com/moby/buildkit#export-cache)) speeding up builds.

derlaft:
> API requests distributed over multiple stateless threads based on available CPUs

So I have tried that in in k8s it might not be productive. Currently in the k8s setup I observe the following:

* Each pod is allocated a small number of resources (~1 cpu core each)
* `HorizontalPodAutoscaler` automatically scales the number of pods based on their load
* Under non-prod load only `~200m` of CPU (m is 1/100 of one logical core) and one pod is more than enough 

With this change:
* Each pod tries to spawn a subprocess for each CPU or close to that (I can see around 18 in the logs) even if it's only allowed to use one-two by cgroups limit
* Because of this, pod resource limit is quickly reached
* Because of this, `HorizontalPodAutoscaler` quickly goes all the way up to the maximum number of pod replicas
* Total CPU usage is `6000m`. `HorizontalPodAutoscaler` can obviously be tweaked, but I don't like the idea that each pod tries to do auto-scaling that k8s is supposed to do.

So:
* in the current form this does not look to be helpful/productive (at least for k8s-based and similar deployments)
* can this new behavior somehow be reconfigured?

lmangani:
Thanks @derlaft for donating time and testing. Those are really useful observations! Clearly this is counter-productive in k8s with autoscaling (would love it if you could share some configuration examples for other users!)

Let me see if pm2 accepts ENV overrides for the type, so it can be set back to fork mode for regular "limited" executions.

PR #117, opened 2022-02-10
Use patched modules

lmangani:
Use patched modules in @qxip npm scope including security fixes unavailable on upstream for plugnplay, copy, to-file modules


Dletta:
Ready, no more vulnerabilities when running npm install

Dletta (approved):

PR #118, opened 2022-02-11
samplesTableName logic

akvlad:
smart choose samples table to read from


Dletta (approved):
Queries in DEV dashboard execute well, detection was successful for old table

Dletta:
Added Master back in, tested locally, ran tests successful read and write

PR #121, opened 2022-02-17
Nanoseconds support

akvlad:
# Changelog
- moved to nanoseconds as timestamp value
- added `samples_v3` as the new table to store the data
- added codecs to `timestamp` and `value` fields
- added a set of views to reach old data 
- moved maintenance scripts into a separate module

# Maintenance scripts
- `lib/db/maintain/scripts.js` is the file to store the sql commands to update the schema
## NOTE!!!
- **!!!append only!!!** no delete or edit after git push.
- Remember that the system is not transactional. `IF NOT EXISTS` is a must have statement.


akvlad:
# Backwards Compatibility
All the readers and writers should be updated all at once. Working with multiple writers to the same database may lead to data loss.
Downgrade to the previous version is not recommended.

akvlad:
# Perfofmance tuning and finishing migration v1 -> v2
- run cloki v2 for the first time to upgrade the schema.
- log in to the clickhouse db
- use your database `use YOUR_DB_NAME`
- copy data from v1 / v2 to the v3 table.
`INSERT INTO samples_v3 (fingerprint, timestamp_ms, value, string) SELECT fingerprint, timestamp_ms * 1000000 as timestamp_ms, value, string from samples_v2;`
`INSERT INTO samples_v3 (fingerprint, timestamp_ms, value, string) SELECT fingerprint, timestamp_ms * 1000000 as timestamp_ms, value, string from samples;`
- drop old samples tables or rename them:
`DROP TABLE samples`
`DROP TABLE samples_v2`

Dletta:
Added Migration to Wiki

akvlad:
Renamed new column to `timestamp_ns`. 
@Dletta and any people installed this branch by accident:
In order to update you should:
- drop table `samples_v3`. Beware! There can be some valuable data!
`DROP TABLE samples_v3`
- drop table samples_read_v2_1
`DROP TABLE samples_read_v2_1`
- drop table `ver`
`DROP TABLE ver`
- restart cLoki

Dletta (approved):

Dletta:
Tested, BREAKING CHANGES as OUTLINED in changelog

PR #129, opened 2022-03-16
watchers reuse

akvlad:


Dletta:
To test:

Open multiple live queries via Grafana and check tables that are generated for watching logs.
Table names start with watcher_....

Dletta (approved):

Dletta:
Tested, as expected.

No new table is generated for LIVE watchers with the same exact query.
No issues found when closing out directly, or stopping. Watcher Tables were correctly removed, when no longer needed.

PR #157, opened 2022-05-11
Tempo fixes

lmangani:
- [x] fix ingestion of clickhouse generated spans w/o content-lenght
- [x] parse tags as attributes in Tempo responses

### DIFF
--- a/cloki.js
+++ b/cloki.js
@@ -116,6 +116,7 @@ async function getContentBody (req) {
  */
 async function genericJSONParser (req) {
   try {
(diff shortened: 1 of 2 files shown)


cloki.js
```diff
@@ -116,6 +116,7 @@ async function getContentBody (req) {
  */
 async function genericJSONParser (req) {
   try {
+    if (req.routerPath === '/tempo/api/push') { return JSON.parse(await getContentBody(req)) } /* deal with bad spans w/o content-length */
```
lmangani, line 119: @akvlad is this an acceptable way to bypass the content size control for broken agents? (clickhouse URL engine)

lmangani: any suggestion here?

PR #176, opened 2022-06-29
grafana 9 support

akvlad:


lmangani (approved):

lib/db/clickhouse.js
```diff
@@ -268,6 +268,7 @@ const queryFingerprintsScan = async function (query, res) {
   // logger.info(_query.query);
   const _stream = await getClickhouseStream(query)
   const dataStream = preprocessStream(_stream, query.stream || [])
+  console.log(query)
```
lmangani, line 271: is this needed?

lmangani: trimmed out

PR #216, opened 2022-09-15
configurable primary key for the samples table

akvlad:
Please specify the `ADVANCED_SAMPLES_ORDERING` env var with the raw clickhouse ORDER BY (...) experssion.
If you want use fingerprint, please specify `ADVANCED_SAMPLES_ORDERING=fingerprint,timestamp_ns` . 
This env variable cannot be changed for the same database.
If you specified this configuration, no samples primary key updates will be applied for the database in the future.
Please use it with care.

### DIFF
--- a/common.js
+++ b/common.js
@@ -82,3 +82,18 @@ module.exports.durationToNs = (durationStr) => {
 module.exports.LineFmtOption = () => process.env.LINE_FMT || 'handlebars'
 
 module.exports.errors = require('./lib/handlers/errors')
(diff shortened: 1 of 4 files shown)


lmangani:
@akvlad shall we merge and document this?

akvlad:
@lmangani yes. We have to document this.

lmangani:
Merge conflicts. Should we refactor or redesign based on the latest developments?

Issue #225, opened 2022-10-13
Unexpected request content-type


Cluas:
{"level":50,"time":1665662651531,"pid":19,"hostname":"qryn-59f4b87599-z9jp9","name":"qryn","reqId":"req-1bn","err":"Unexpected request content-type\nError: Unexpected request content-type\n    at genericJSONOrYAMLParser (/app/qryn.js:210:11)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async Parser.fn (/app/qryn.js:373:12)","msg":"Unexpected request content-type"}
{"level":30,"time":1665662651531,"pid":19,"hostname":"qryn-59f4b87599-z9jp9","name":"qryn","reqId":"req-1bn","res":{"statusCode":200},"responseTime":6.672591000795364,"msg":"request completed"}

akvlad:
Hello @Cluas, which client do you use?
Can you check if the content-type of your request is exactly `application/json`?

Cluas:
grafana agent config:
 ```yaml
  agent.yaml: |-
    server:
      log_level: debug

    traces:
      configs:
      - name: server_traces
        receivers:
          otlp:
            protocols:
              http:
                endpoint: "0.0.0.0:4318"
              grpc:
                endpoint: "0.0.0.0:4317"
        remote_write:
          - endpoint: qryn:3100
            insecure: true
```

Cluas:
grafana agent logs:
``` bash
ts=2022-10-13T12:13:01Z level=error caller=exporterhelper/queued_retry.go:183 msg="Exporting failed. The error is not retryable. Dropping data." component=traces traces_config=server_traces kind=exporter data_type=traces name=otlp/0 error="Permanent error: rpc error: code = Unknown desc = transport: received unexpected content-type \"application/json; charset=utf-8\"" dropped_items=121
ts=2022-10-13T12:13:03.131212936Z caller=config_watcher.go:139 level=debug agent=prometheus component=cluster msg="waiting for next reshard interval" last_reshard=2022-10-13T12:13:03.131182084Z next_reshard=2022-10-13T12:14:03.131182084Z remaining=59.999995185s
[...]

PR #227, opened 2022-10-14
fix annotations

akvlad:
fix tempo annotations value


akvlad:
as for https://github.com/metrico/qryn/issues/225

PR #236, opened 2022-11-03
Csv output

akvlad:
add `&csv=1` to `query_range` to get csv response.


lmangani:
@SocketSecurity ignore qryn@2.1.39

PR #267, opened 2022-12-03
Promql non-handlers

lmangani:
The PromQL scope is limited and certain commands should be excluded from the transpiler.
Example: `node_info` seems to be used as ping by a few prometheus drivers and fails validation.

Certain APIs were only implemented with Grafana in mind. Missing routes/methods should be backfilled.


lmangani:
@akvlad do you have a better idea? The call seems to return all active labels in the cloud api

akvlad:
seems ok

PR #284, opened 2022-12-30
Json tempo output fix

akvlad:
json output of temo fix


akvlad:
test passed. merging.

Issue #288, opened 2023-01-26
Prometheus metrics ingestion: snappy compression issue (Fastify)


inatale:
Hello team!
Thank for your product. We are testing it right now for gathering logs and metrics. We have successfully configured logs ingestion using Vector agent and your Loki inteface.

But we have an issue with metrics ingestion. We send Prometheus metrics but receive an error from remote write:
Unsupported Content-Encoding: snappy
FastifyCompressError: Unsupported Content-Encoding: snappy
    at Object.preParsing (/app/node_modules/@fastify/compress/index.js:315:24)
    at next (/app/node_modules/fastify/lib/route.js:552:16)
    at preParsingHookRunner (/app/node_modules/fastify/lib/route.js:575:3)
    at runPreParsing (/app/node_modules/fastify/lib/route.js:526:5)
    at next (/app/node_modules/fastify/lib/hooks.js:168:7)
    at Object.onRequest (/app/node_modules/@fastify/compress/index.js:205:5)
    at hookIterator (/app/node_modules/fastify/lib/hooks.js:246:10)
    at next (/app/node_modules/fastify/lib/hooks.js:174:16)
    at /app/node_modules/@fastify/cors/index.js:130:12
    at resolveOriginOption (/app/node_modules/@fastify/cors/index.js:88:129)

We have investigated it and found out that the issue is related to Fastify library (probably 4.10 version). Fastify doesn't support snappy compression which is used sending packages. We could mitigate the issue falling down to 2.1.56 image tag. In the next to 2.1.56 release you have stated that change something with compression (that is related to the issue).

lmangani:
Thanks for the report @inatale we're looking into this and will have an update shortly. 

FYI and to save you precious time in case this is relevant: qryn-js happily ingests any metrics in many formats, but mind it currently only offers a promql-to-logql transpiler with a few limitations, until our flux integration is completed to swap it out. Full PromQL query support is currently available on qryn.cloud already.

akvlad:
@inatale hello.
Should be fixed in v2.2.3. Please check.

inatale:
@akvlad thank you. Now it works. Only tag is not v2.2.4 but v2.2.6.

PR #289, opened 2023-01-26
Content encoding snappy fix

akvlad:
https://github.com/metrico/qryn/issues/288 fix


akvlad:
Minor changes
Tests ok.
Merging

Issue #290, opened 2023-02-01
ClickHouse Materialized View + URL engine send to /influx/api/v2/write  return "404 Not Supported"


gohonsen:
I refer to this link: [https://qryn.metrico.in/#/logs/ingestion ](url) ,  after create ClickHouse Materialized View + URL engine, but qryn view show nothing

I follow this example:

> `CREATE MATERIALIZED VIEW default.influx_log_send
ENGINE = URL('http://qryn:3100/influx/api/v2/write', 'LineAsString')
AS SELECT format('syslog,level={0},logger_name={1} message="{2}" {3}', 
  toString(level), 
  replaceRegexpAll(toString(logger_name), '[^a-zA-Z0-9_]', '_'), 
  replaceAll(replaceRegexpAll(message, '["\\\\]', '\x00\\0'), '\x00', '\\'), 
  toString(toUnixTimestamp64Nano(event_time_microseconds))) FROM system.text_log;
`
>

create successfully, then I select from default.influx_log_send return "404 Not Supported" , and I access the /influx/api/v2/write url by browser, return "404 Not Supported",  too.


SELECT *
FROM default.influx_log_send

 ┌─format('syslog,level={0},logger_name={1},type=clickhouse message="{2}" {3}', toString(level), replaceRegexpAll(toString(logger_name), '[^a-zA-Z0-9_]', '_'), replaceAll(replaceRegexpAll(message, '["\\\\]', '\0\\0'), '\0', '\\'), toString(toUnixTimestamp64Nano(event_time_microseconds)))─┐
  │ 404 Not Supported                                                                                                                                                                                                                                          
  │
 └───────────────────────────────────────────────────────

ClickHouse version is 23.1.1.3077
Qryn use docker image qxip/qryn:v2.2.6

lmangani:
Hello @gohonsen could you confirm the version of qryn and clickhouse in use?

lmangani:
Mind this is a POST endpoint, it won't support GET requests from your browser:
```
fastify.post('/influx/api/v2/write', handlerInfluxWrite, {
    '*': rawStringParser
  })
```

gohonsen:
> Hello @gohonsen could you confirm the version of qryn and clickhouse in use?
ClickHouse version is 23.1.1.3077
Qryn version is v2.2.6

akvlad:
@gohonsen hello.
There's a branch for you. https://github.com/metrico/qryn/tree/influx_fix_290
[...]

PR #291, opened 2023-02-01
influx parser fix

akvlad:
Fix https://github.com/metrico/qryn/issues/290


akvlad:
tests ok.
local check ok.
Merging for now.

PR #296, opened 2023-02-17
more e2e test merge fixes

akvlad:
- elastic header
- influx response code


lmangani:
closing just to keep things clean

Issue #297, opened 2023-02-21
Please help to check a write performance issue


zhushilu:
I had met a performance issue about the insert rate of qryn. As my testing,the insert rate is ～5K rows per second, have a huge gap between your description of 500K per second, so I have a question about how to improve my write rate enormously, my testing result and screenshot as below:

my  qryn docker-compose.yml:
```
version: '3'
services:
  qryn:
      image: qxip/qryn:latest
      ports:
        - "13100:3100"
      environment:
        - NODE_OPTIONS="--max-old-space-size=8192"
        - FASTIFY_METRICS=true
        - CLICKHOUSE_SERVER=172.30.2.100
        - CLICKHOUSE_AUTH=default:123456
        - CLICKHOUSE_PROTO=http
        - CLICKHOUSE_PORT=8133
        - CLICKHOUSE_DB=default
```

clickhouse docker-compose.yml
```
version: '3'

services:
  clickhouse:
    image: clickhouse/clickhouse-server:22.10.2.11
    hostname: clickhouse
    container_name: clickhouse
    expose:
      - "9000"
    user: root
    privileged: true
    ports:
      - "8133:8123"
      - "9011:9000"
    volumes:
      - ./data/cloki:/var/lib/clickhouse
      - ./log/:/var/log/clickhouse-server/
      - ./etc/:/etc/clickhouse-server/
```

client configfile:
```
server:
  http_listen_port: 0
  grpc_listen_port: 0

limits_config:
  readline_rate: 1000000
  readline_burst: 1800000
  readline_rate_drop: true
  #max_streams: 100
  readline_rate_enabled: false

positions:
  filename: /opt/positions-kafka.yaml 
  sync_period: 10s

clients:
- url: http://172.30.2.100:13100/loki/api/v1/push
  batchwait: 2s
  batchsize: 10485760
  external_labels:
    cluster: pre

scrape_configs:
- job_name: kafka
  kafka:
    brokers:
    - 172.30.2.110:9092
    - 172.30.2.111:9092
    - 172.30.2.112:9092
    - 172.30.2.113:9092
    - 172.30.2.114:9092
    topics:
    - pre
    group_id: "pre"
    version: "0.10.2.0"
    labels:
      topic: pre
    use_incoming_timestamp: true
  pipeline_stages:
  - json:
      expressions:
        app_name: type
  - labels:
      level:
      app_name:
  - output:
      source: output
```
kafka monitor:
[...]

PR #298, opened 2023-02-21
perf issue for protobuf

akvlad:
as described in https://github.com/metrico/qryn/issues/297


akvlad:
minor fix
tests passed ok.
merging

Issue #312, opened 2023-03-17
otlp push api gzip decompress error


Cluas:
we use https://www.npmjs.com/package/@fastify/compress

if content-encoding defined, we need read data from payload.

PR #313, opened 2023-03-18
fix: otlp push gzip decompress error

Cluas:
https://www.npmjs.com/package/@fastify/compress will change payload to zlib.Gunzip

we need transform data from payload to body.

this will solve https://github.com/metrico/qryn/issues/312


akvlad:
I'll PR the branch into our ``fix-otlp_push` branch . I can't do git pull now.

PR #314, opened 2023-03-20
Fix otlp push

akvlad:
Merge this after a retest https://github.com/metrico/qryn/pull/313
***
https://www.npmjs.com/package/@fastify/compress will change payload to zlib.Gunzip

we need transform data from payload to body.

this will solve https://github.com/metrico/qryn/issues/312


akvlad:
Tests passed. Backwards compatibility should be ok.

PR #315, opened 2023-03-20
fix: otlp status missing

Cluas:


akvlad:
Merging to a dedicated branch for some investigation.

PR #342, opened 2023-08-11
fix: logfmt should produce 0 or more extra labels

akvlad:
Example:
line: `INFO: <script>: Reply from Inbound - S=123 - Proxy Authentication Required M=INVITE IP=udp:127.0.0.1:1234 ID=abcd@127.0.0.1`
parse result: 
```
{
  S: '123',
  M: 'INVITE',
  IP: 'udp:127.0.0.1:1234',
  ID: 'abcd@127.0.0.1'
}
```


akvlad:
e2e passed successfully. Minor update should work ok.

PR #347, opened 2023-09-23
Feat/pyroscope integration

tomershafir:


lmangani:
Q: I don't see anything related to ingestion or how the data should be inserted into the new tables or API mapping. Are there more files missing, or what was the testing method to develop? Thanks @tomershafir

akvlad:
I see no handlers in the PR. How do we ingest the data?

lmangani:
We might have to wait for @tomershafir to return from Japan to complete this integration

lmangani:
New Plan:

- MVP write integration in qryn-js for `pprof` ingestion to the proposed CH schema
- MVP read integration in qryn-js to serve Grafana Pyroscope API off the proposed CH schema
- Opentelemetry qryn-otel-collector extensions to ingest pprof, jfr to the proposed CH schema

PR #361, opened 2023-10-22
3.0: Always use dbname

lmangani:
- Explicitly use the database name in each query, limit assumed database names or `USE` statements


bzed:
@akvlad a review and merge would be apprciated, working around this issue is rather annoying

akvlad:
@bzed ok.Will merge tomorrow. Thanks for the upvote.

akvlad (approved):

akvlad:
@bzed reviewed and merged.

Issue #369, opened 2023-10-30
qryn generates 500 error when auth is missing


tsearle:
`{
"level": 50,
"time": 1698681199146,
"pid": 18,
"hostname": "[ip-10-150-4-38.eu](http://ip-10-150-4-38.eu/)-west-1.compute.internal",
"name": "qryn",
"reqId": "req-5w",
"err": "Missing or bad formatted authorization header\nUnauthorizedError: Missing or bad formatted authorization header\n    at Object.basicAuth (/app/node_modules/fastify-basic-auth/index.js:23:12)\n    at hookIterator (/app/node_modules/fastify/lib/hooks.js:246:10)\n    at next (/app/node_modules/fastify/lib/hooks.js:174:16)\n    at hookRunner (/app/node_modules/fastify/lib/hooks.js:196:3)\n    at preValidationCallback (/app/node_modules/fastify/lib/handleRequest.js:104:5)\n    at handler (/app/node_modules/fastify/lib/handleRequest.js:76:7)\n    at handleRequest (/app/node_modules/fastify/lib/handleRequest.js:24:5)\n    at next (/app/node_modules/fastify/lib/route.js:545:7)\n    at Object.preParsing (/app/node_modules/[@fastify/compress/index.js:299:14](https://matrix.to/#/@fastify/compress/index.js:299:14))\n    at next (/app/node_modules/fastify/lib/route.js:552:16)",
"msg": "Missing or bad formatted authorization header"
}`

this sending a 500 instead of a 401 prevents the browser from generating the user/pass popup

this means the UI is not usable when QRYN_USER and QRYN_PASSWORD are set

akvlad:
https://github.com/metrico/qryn/pull/371

lmangani:
@tsearle release 3.0.1 should provide a fix for this issue. Would you have time to retest and let us know next time around?

tsearle:
![Screenshot 2023-11-03 at 10 55 44](https://github.com/metrico/qryn/assets/4540323/7aa254d1-49be-4293-b2ea-c461aae07b35)

it is working :-) thanks

PR #371, opened 2023-10-31
Basic auth fix 369

akvlad:
- basic auth 401 response fix
- CORS for bun
- qryn-view  serve for bun
- `content-type: application/json` minor bug fix


akvlad:
#369

Issue #390, opened 2023-11-15
occasional exception in CPU graph


tsearle:
using a slightly customised version of this dashboard
https://grafana.com/grafana/dashboards/19650-node-exporter-vector-host-metrics/

somtimes the cpu graph errors with this exception, sometimes it does not

Nov 15 15:11:02 ip-10-150-3-194 qryn[587]: Empty <[Object: null prototype] {}> {
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:   end: '1700061060',
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:   query: 'sum by(instance) (irate(host_cpu_seconds_total{instance="ip-10-150-4-35", mode="idle"}[1m0s])) / on(in>
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:   start: '1700060760',
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:   step: '15'
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]: }
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]: RangeError: offset is out of bounds
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at Uint8Array.set (<anonymous>)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at Uint8ArrayWriter.writeString (/usr/lib/node_modules/qryn/wasm_parts/main.js:274:14)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at pql (/usr/lib/node_modules/qryn/wasm_parts/main.js:167:14)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at async module.exports.pqlRangeQuery (/usr/lib/node_modules/qryn/wasm_parts/main.js:58:10)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at async module.exports.rangeQuery (/usr/lib/node_modules/qryn/promql/index.js:20:18)
Nov 15 15:11:02 ip-10-150-3-194 qryn[587]:     at async Object.handler (/usr/lib/node_modules/qryn/lib/handlers/prom_query_range.js:22:20)

akvlad:
should be fixed. Merged to master. @lmangani if you have time, please release. Or I'll do it later.

PR #391, opened 2023-11-15
Fix out of bounds

tsearle:
proposed fix for #390


akvlad:
Right. Thanks a lot for the fix.

PR #417, opened 2023-12-22
fix: rm fixed allocation for unused profile id

tomershafir:


akvlad:
profile_id fileld was dropped in 
https://github.com/metrico/qryn/commit/bf95d2d6d7544fa343878274f87bb1c4937629b4

PR #456, opened 2024-02-05
profiles: add ttl to clickhouse tables

tomershafir:


akvlad:
@tomershafir can you please fix the conflict?

PR #514, opened 2024-06-12
Release 3.2.23 prep

akvlad:
- #513 
- #512
- #510
- #508 

### Qryn operation Mode

MODE env var was introduced. It can have one of 3 values:
- `writer` - only writing paths
- `reader` - only reading paths
- `init_only` - initialize the database and exit

If the env var is omitted, AIO mode is applied as it was before.

### Bulk size configuration

Two env vars were introduced:
- `BULK_MAX_SIZE_BYTES` - max size of the bulk in bytes (default - infinite)
- `BULK_MAX_AGE_MS` - max age of the bulk in milliseconds (default - 100), can't be less than 100


lmangani:
[Relevant Docs](https://github.com/metrico/qryn-docs/blob/main/docs/env.md) updated w/ new parameters

Issue #556, opened 2024-08-29
Explore profiles plugin fails on labels/label values query


tomershafir:
qryn 3.2.8
plugin 0.1.6
grafana 10.4.7

<img width="1728" alt="Screenshot 2024-08-29 at 11 40 19" src="https://github.com/user-attachments/assets/0b5828f8-96b5-4a61-aec4-c58bdbbe46dd">
<img width="1728" alt="Screenshot 2024-08-29 at 11 41 42" src="https://github.com/user-attachments/assets/5edb7a69-9a99-425b-a24d-7878087dcf97">

Cluas:
@tomershafir this PR https://github.com/metrico/qryn/pull/557 will fix it.

tomershafir:
Works on
qryn 2.2.32
plugin 0.1.8
grafana v11.2.0

PR #557, opened 2024-08-29
feat: labelValues support json format

Cluas:
grafana Profiles use application/json as content-type


Cluas:
fix https://github.com/metrico/qryn/issues/556

lmangani (approved):

PR #589, opened 2024-10-14
feat: support insert trace through the distributed table

Cluas:
Add `traces_input_dist` table, so we can insert data through the distributed table.


Cluas:
Using non-traces_input and dispersing the pressure of ClickHouse is a good solution.

Issue #602, opened 2024-12-12
Feature Request: Support legacy Elastic _bulk API clients which use `Content-Type: application/json`


awesense-paul:
##### Background
- The Elastic `_bulk` API is effectively multiple JSON objects concatenated one after the other; separated by newlines.
- Older Elasticsearch clients (e.g. rsyslog, others) of the `_bulk` API may specify a legacy/incorrect `Content-Type` header of `application/json` (this implies a single JSON document). The newer -- correct -- `Content-Type` is "`Content-Type: application/x-ndjson`", but since this is a newer standard, the old clients never implemented using it.
- qryn allows only `application/x-ndjson`, and rejects requests which use `application/json`.

##### Change Request
If qryn permitted "`Content-Type: application/json`", this would allow the older clients to be supported without changes. (Even though this is not strictly correct in the current time.) Perhaps it could be a setting whether to permit it or not?

awesense-paul:
This was the change that I found worked for me:

https://github.com/awesense-paul/qryn/commit/9ad456eb3f7650edb2728cb492fbe1674854f4fb

(the mentioned code block actually had the same body in both the `if` and `else`. I think there's a better fix that actually uses `_rawBody` in the `else` block, but tbh I don't really understand the codebase+fastify well enough to know what to do.)

awesense-paul:
> I think there's a better fix

(or not, since the later handling does a [JSON.parse()](https://github.com/metrico/qryn/blob/master/lib/handlers/elastic_bulk.js#L48) on each element in `streams` anyway)

lmangani:
@awesense-paul please send in a PR and we'll adjust and merge

awesense-paul:
This issue appears to be resolved / non-issue in `gigapipe`

PR #606, opened 2025-01-07
Support legacy Elastic _bulk API clients which use `Content-Type: application/json`

awesense-paul:
Changes the Elastic bulk API parser to support requests with `Content-Type: application/json`, by sending all requests through the same code path. (the existing parser behaviour expected a single JSON object for `application/json` -- though that code path would error anyway -- which is still supported as the case where there is 1 object / no newlines).

Resolves https://github.com/metrico/qryn/issues/602


awesense-paul:
(sorry, it was not my intention to close the pull request -- I was only intending to rebase on latest `master` and GPG-sign my commit)

lmangani:
FYI @awesense-paul the next version of qryn (aka gigapipe) is now in a branch on this repo. The new codebase is in golang so this change would have to be ported there to be effective in the next release.

lmangani:

PR #618, opened 2025-03-05
Remove some noisy `console.log` statements

awesense-paul:
Removes quite a few lines that end up in log files (which also then end up in ClickHouse if logs are forwarded for `qryn`).


awesense-paul:
(These could be made `console.debug` statements instead of being removed, if they are still useful statements to have for `qryn` development)

lmangani:

PR #619, opened 2025-03-05
Set "Content-Type: application/proto" for pyroscope `QuerierService` responses

awesense-paul:
Latest Grafana (11.5.2 at time of writing this) appears to require "`Content-Type: application/proto`" for responses from pyroscope `QuerierService`. (previously, "`Content-Type: application/octet-stream`" was used: the fastify default).


lmangani:

PR #653, opened 2025-04-22
Vector implements

afzal-qxip:
Vector implements a health check method by sending a get to the /prom/remote/write

Screenshot 2025-04-22 at 5 39 45 PM


lmangani:
LGTM

PR #660, opened 2025-05-01
Fix issue with "snappy: corrupt input"

VolanDeVovan:
I encountered a problem where the Prometheus push endpoint wasn't working and returned the following error:
{"level":"error","msg":"snappy: corrupt input","time":"2025-05-01T12:43:09Z"}

I tested this using Alloy and Promtool.
After some research, I discovered that the problem occurred because it handled "snappy" multiple times.
Commenting out these lines fixed the issue.
However, I have only tested this with Prometheus, and I'm not sure if it might break other functionality. Please review and test.

### DIFF
--- a/writer/controller/middleware.go
+++ b/writer/controller/middleware.go
@@ -6,15 +6,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/golang/snappy"
-	"github.com/metrico/qryn/writer/ch_wrapper"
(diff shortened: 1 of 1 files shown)


akvlad:
Thanks for the update. I'll retest.

ggontariu:
We fixed this issue by changing the compression settings in vector to gzip

https://github.com/metrico/gigapipe/issues/650

PR #674, opened 2025-06-05
Fixed incorrect key parameter passing to function os.Getenv

vershinink:
In the ```boolEnv``` function, the key parameter is passed to ```os.Getenv()``` in double quotes, which means that a string value is passed instead of a key variable. So the OMIT_CREATE_TABLES environment variable is not used.


akvlad:
@vershinink great update. thanks a lot for the fix.

PR #678, opened 2025-06-10
timestamp fix

afzal-qxip:


rita7lopes:
Applied in https://github.com/metrico/gigapipe/pull/829

Issue #744, opened 2025-10-29
LOG_LEVEL environment variable is ignored — logs remain at info level even when set to warn


TipTopBY:
# Description

When running Gigapipe in Kubernetes, the container ignores the LOG_LEVEL environment variable and continues logging at the info level, even though the environment is explicitly configured with LOG_LEVEL=warn

Gigapipe Version: 4.0.22

# Environment details

**Pod details:**
```
Containers:
  gigapipe:
    Container ID:   containerd://7bef9c26a6fe6c182ed67d0037c77b1a1292a534ff0a263a9ba3bd37231928f5
    Image:          XXXXXXXXXXXXXXXXXXXXXXXX
    Image ID:       XXXXXXXXXXXXXXXXXXXXXXXXXX
    Port:           3100/TCP
    Host Port:      0/TCP
    State:          Running
      Started:      Wed, 29 Oct 2025 11:57:34 +0300
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     3
      memory:  10Gi
    Requests:
      cpu:     100m
      memory:  128Mi
    Liveness:  http-get http://:http/metrics delay=10s timeout=1s period=10s #success=1 #failure=3
    Environment:
      CLICKHOUSE_AUTH:    gigpipe:XXXXXXXXXXXX
      CLICKHOUSE_DB:      gigapipe
      CLICKHOUSE_PORT:    9000
      CLICKHOUSE_SERVER:  clickhouse-cluster01.clickhouse.svc.cluster.local
      LABELS_DAYS:        5
      LOG_DRILLDOWN:      true
      LOG_LEVEL:          warn
      SAMPLES_DAYS:       5
      STORAGE_POLICY:     s3_cache_tiered
    Mounts:               <none>
Conditions:
  Type                        Status
  PodReadyToStartContainers   True 
  Initialized                 True 
  Ready                       True 
  ContainersReady             True 
  PodScheduled                True 
Volumes:                      <none>
QoS Class:                    Burstable
Node-Selectors:               <none>
Tolerations:                  node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                              node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:                       <none>
```

**Variables inside pod:**
```
/ # env
LOG_DRILLDOWN=true
KUBERNETES_PORT=tcp://10.43.0.1:443
KUBERNETES_SERVICE_PORT=443
STORAGE_POLICY=s3_cache_tiered
[...]

PR #745, opened 2025-10-31
fix: issue #744 - updated logging behavior

44YHC:


akvlad:
Unfortunately there is a lot of potentially descructive changes for a minor request. We have to reconsoder the approach.

PR #767, opened 2026-01-19
Fix: Extract match[] parameter from URL query for GET requests in labels endpoint

uasy:
# Fix: Extract match[] parameter from URL query for GET requests in labels endpoint

## Problem

When querying the `/api/v1/labels` endpoint with a `match[]` parameter for metric filtering via GET request, the filtering was not applied. All labels were returned without considering the metric filter.

### Example of problematic request

```
GET /api/v1/labels?limit=40000&start=1768818540&end=1768822200&match[]={__name__=~".*process_cpu_utilization.*"}
```

**Expected behavior:** Only labels associated with metrics matching the filter `{__name__=~".*process_cpu_utilization.*"}` should be returned.

**Actual behavior:** All labels were returned without applying the filter.

## Root Cause

The `getLabelsParams` function in `reader/controller/prom_query_labels.go` was only extracting the `match[]` parameter for POST requests, but not for GET requests. As a result, for GET requests `params.match` was `nil`, and filtering was not applied in the `QueryLabelsService.Labels` method.

### Code before fix

```go
func getLabelsParams(r *http.Request) (*promLabelsParams, error) {
    if r.Method == "POST" && r.Header.Get("content-type") == "application/x-www-form-urlencoded" {
        // ... correctly extracted match[] from r.Form["match[]"]
        return &promLabelsParams{
            start: parserTimeString(rawParams.Start, time.Now().Add(time.Hour*-6)),
            end:   parserTimeString(rawParams.End, time.Now()),
            match: rawParams.Match,  // ✅ extracted for POST
        }, nil
    }
    
    // ❌ ISSUE: match[] was not extracted for GET requests
    return &promLabelsParams{
        start: parserTimeString(r.URL.Query().Get("start"), time.Now().Add(time.Hour*-6)),
        end:   parserTimeString(r.URL.Query().Get("end"), time.Now().Add(time.Hour*-6)),
        // match was missing - was nil
    }, nil
}
```

## Solution

Added extraction of the `match[]` parameter from URL query string for GET requests, consistent with how other functions (`getPromSeriesParamsV2`, `ParseLogSeriesParamsV2`) handle it.

### Change

```go
return &promLabelsParams{
    start: parserTimeString(r.URL.Query().Get("start"), time.Now().Add(time.Hour*-6)),
    end:   parserTimeString(r.URL.Query().Get("end"), time.Now()),
    match: r.URL.Query()["match[]"],  // ✅ Added match[] extraction for GET
}, nil
```

## Files Changed

- `reader/controller/prom_query_labels.go` - added `match[]` extraction from URL query for GET requests


rita7lopes:
Hello @uasy
PRs without CLA signed are not considered to be reviewed/merged.
Furthermore, Gigapipe has no code forks/copies from other open-source repos.

rita7lopes:
The fix on this PR was applied on https://github.com/metrico/gigapipe/pull/829. You can take version 4.2.3, which has the latest fix in place

PR #812, opened 2026-05-29
fix: remove dead qryn_fingerprints table variable in writer init

rita7lopes:
## Summary

Removes a dead code block in writer plugin init: the `table` variable (set to `qryn_fingerprints` or `qryn_fingerprints_dist`) was computed but never used.

## Test plan
- [ ] Verify writer builds and initialises cleanly


44YHC:
Approved!

PR #864, opened 2026-07-11
fix: order of route registration

### LINKED ISSUE #862: Recording rules API routes unreachable when built with -tags view (VIEW=1) in CI
**Summary:**

The ruler HTTP routes (`/api/prom/rules/{namespace}`, `/api/v1/rules/{namespace}`, `/loki/api/v1/rules/{namespace}`) return 404 when gigapipe is built with `VIEW=1` (the default for published ghcr.io images). The ruler starts successfully and evaluates rules, but the CRUD API is inaccessible.

**Root Cause:**

In `cmd/gigapipe/main.go`, the initialization order is:

```go
reader.Init(cfg, app)
view.Init(cfg, app)       // ← registers PathPrefix("/") catch-all
rulerrouter.Init(cfg, app) // ← ruler routes registered AFTER the catch-all
```

When built with `-tags view`, `view.Init` (`view/view.go`) registers a `PathPrefix("/")` handler for the embedded static file server:

```go
mux.PathPrefix(prefix).Handler(fileServer)  // prefix defaults to "/"
```

In gorilla/mux, `PathPrefix("/")` matches every request. Since it's registered before the ruler routes, it takes priority and the ruler handlers are never reached.

**Reproduction:**

```bash
# ghcr.io images (built with VIEW=1) — ruler routes 404:
docker run --rm -e QRYN_RULER_ENABLED=true -e CLICKHOUSE_SERVER=... ghcr.io/metrico/gigapipe:v4.3.1
curl -X POST http://localhost:3100/api/prom/rules/test  # → 404 page not found

# Local build without VIEW=1 — ruler routes work:
go build -o gigapipe ./cmd/gigapipe
curl -X POST http://localhost:3100/api/prom/rules/test  # → 202 success
```

**Affected versions:** v4.3.0, v4.3.1 (any ghcr.io image built with `VIEW=1`)

**Suggested fix:**

Move `rulerrouter.Init(cfg, app)` before `view.Init(cfg, app)` in `cmd/gigapipe/main.go`:

```go
reader.Init(cfg, app)
rulerrouter.Init(cfg, app)  // ruler routes registered BEFORE the catch-all
view.Init(cfg, app)
```

Or change `view.Init` to use a more specific prefix that doesn't shadow API routes.

mgai-rchung:
The view server registers a default route with the mux, so it has to
come last before the http server is started.

Fixes: #862

### DIFF
--- a/cmd/gigapipe/main.go
+++ b/cmd/gigapipe/main.go
@@ -4,20 +4,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"github.com/gorilla/mux"
-	"github.com/grafana/pyroscope-go"
-	clconfig "github.com/metrico/cloki-config"
-	"github.com/metrico/cloki-config/config"
-	"github.com/metrico/qryn/v4/ctrl"
-	"github.com/metrico/qryn/v4/reader"
-	rulerrouter "github.com/metrico/qryn/v4/ruler/router"
-	"github.com/metrico/qryn/v4/reader/utils/logger"
-	"github.com/metrico/qryn/v4/reader/utils/middleware"
-	"github.com/metrico/qryn/v4/reader/utils/tables"
-	"github.com/metrico/qryn/v4/shared/commonroutes"
-	"github.com/metrico/qryn/v4/shared/distconfig"
-	"github.com/metrico/qryn/v4/view"
-	"github.com/metrico/qryn/v4/writer"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


akvlad:
@mgai-rchung 
Hello and thanks a lot for the nice bugfix.
LGTM.
Will be merged.

rita7lopes:
Hello @mgai-rchung,

Thank you for your contribution! We will merge alpha (where your contribution currently is) to main today.

However, we spotted something worth mentioning just FYI: there are two competing ordering constraints on the code.
You correctly spotted that view.Init registers a wildcard mux.PathPrefix("/") catch-all that shadows anything registered after it, so it has to come last. That part is right.

But moving reader.Init and view.Init together to after the ruler reintroduces a different, quieter bug: ruler.Init captures session := readerregistry.Registry, and that registry is populated by reader.Init. With reader now running after the ruler, the ruler binds its rule sessions to a nil registry. 
It doesn't break at startup, only at rule-evaluation time, which is why it looked like it was working in testing (no querying rules loaded).

The two constraints:
- reader before ruler — ruler binds to reader's registry (nil session otherwise)
- view last — its wildcard route shadows everything after it

Only reader.Init has no wildcard; the catch-all is purely in view.Init. So the fix is to split them rather than move them as a pair: writer → reader → ruler → view. 
reader ahead of the ruler, view still dead last. That's the one order satisfying both.

This is all sorted on alpha in #874, which also adds a regression test that pins both cases (feeds both bad orderings through the same invariant check so neither can come back silently). 

No action needed here. Just flagging so the reasoning is on record.

PR #868, opened 2026-07-13
Add support for graceful shutdown

yonatan-xendit:
### Summary

Gigapipe currently has no signal handling. When the container runtime sends SIGTERM (e.g. during a rolling deploy), the process dies immediately — dropping in-flight HTTP requests and losing any data still buffered in the writer's batch pipeline.

This PR wires up `SIGTERM`/`SIGINT` handling in `main.go` and fixes both Dockerfiles so the signal actually reaches the process.

### What changed

**`cmd/gigapipe/main.go`**
- Trap `SIGTERM` and `SIGINT` using `signal.Notify`
- On signal, execute an ordered shutdown sequence:
  1. `httpServer.Shutdown()` — stop accepting new connections, drain in-flight requests (30s timeout)
  2. `rulerrouter.Stop()` — cancel evaluation loops, wait for in-flight rule evaluations
  3. `writer.Stop()` — flush batch buffers to ClickHouse, close DB connections
  4. `reader.Stop()` — stop watchdog and registry
- Replaced bare `http.Serve` (blocking) with `http.Server` + goroutine so the main goroutine can block on the signal channel

**`Dockerfile` and `scripts/deploy/docker/Dockerfile`**
- Changed `CMD /gigapipe` (shell form) → `ENTRYPOINT ["/gigapipe"]` (exec form) so the binary is PID 1 and receives signals directly
- Added `STOPSIGNAL SIGTERM` for explicitness

### Why this order

| Step | Reason |
|------|--------|
| HTTP shutdown first | Stop accepting new work before tearing down backends |
| Ruler before writer | Ruler writes back evaluation results via the writer pipeline; stop producing before flushing |
| Writer before reader | Writer flushes buffered inserts to ClickHouse; reader has no buffered state |

### Risk

- Low. The existing `Stop()` functions on writer, reader, and ruler were already implemented and well-structured — they just weren't being called.
- The 30s shutdown timeout aligns with Kubernetes' default `terminationGracePeriodSeconds`.
- No behavior change during normal operation; the signal handler only activates on termination.

### Testing

- `go vet ./cmd/gigapipe/` — passes
- `go test ./...` — all tests pass (one pre-existing `TestTranspilerV2` nil-pointer failure unrelated to this change)
- `make e2e-build` + `make e2e-test` — 159/160 pass (one pre-existing flaky websocket test unrelated to this change)


akvlad:
@yonatan-xendit 
Hello! And thanks a lot for a nice PR. 
I don't really like the `ENTRYPOINT` operator. It collides with `docker exec`. It will be changed.
The rest is LGTM. Thanks a lot.

akvlad:
@yonatan-xendit 
Apparently, regression says that something wrong.
You can run the e2e test suite with `make e2e-full` . The only dependency for that is `docker` and `git`. Please fix the regressions before we can merge.

yonatan-xendit:
Hi, thanks for the reply @akvlad

Reverted back to `CMD` and `make e2e-full` passes 100% (2 times out of 2).

The initial run with `ENTRYPOINT` had a flaky websocket test (`should ws`) that failed once then passed on retry — likely a timing race unrelated to the change. After switching to `CMD ["/gigapipe"]`, I ran the full suite twice and both returned all green:

```
Test Suites: 1 passed, 1 total
Tests:       160 passed, 160 total
Snapshots:   120 passed, 120 total
Time:        44.253 s
Ran all test suites.
```

rita7lopes:
Hello @yonatan-xendit,

Thank you very much for this enhancement, we strongly appreciate it!

I've merged master into your branch to clear the conflict. The conflict came from #874 landing on master after this PR opened. Nothing more.

We are going to merge this PR. Thank you again!

rita7lopes (approved):

PR #886, opened 2026-08-03
fix(reader): detect epoch magnitude when parsing time parameters

### LINKED ISSUE #885: panic: makeslice: len out of range on Loki index/volume with nanosecond timestamps
### What happens

Any Loki API request that sends `start`/`end` as **nanosecond** epochs makes the
process do a full table scan and then die with:

```
panic: makeslice: len out of range
  reader/plannerV2.(*FixPeriodPlanner).Process.func2
```

This is reachable from stock Grafana: enabling **Grafana Logs Drilldown**
(`grafana-lokiexplore-app`) against a Gigapipe Loki datasource crashes the
Gigapipe pod on page load, because the plugin sends nanosecond timestamps.

### Reproduce

Against a Gigapipe instance with a small amount of log data:

```bash
# nanosecond epochs (what Grafana Logs Drilldown sends) -> 200 after 39.2s, then the process dies
curl -s -o /dev/null -w '%{http_code} %{time_total}\n' \
  'http://localhost:3100/loki/api/v1/index/volume?query=%7Bhost_name%3D~%22.%2B%22%7D&start=1785416892000000000&end=1785417792000000000&limit=3'

# same instant, second precision -> 200 in 0.37s
curl -s -o /dev/null -w '%{http_code} %{time_total}\n' \
  'http://localhost:3100/loki/api/v1/index/volume?query=%7Bhost_name%3D~%22.%2B%22%7D&start=1785416892&end=1785417792&limit=3'
```

### Cause

`ParseTimeSecOrRFC` in `reader/controller/utils.go` treats every numeric time
parameter as **seconds**:

```go
t, err := strconv.ParseFloat(raw, 64)
...
return time.Unix(int64(sec), int64(frac*float64(time.Second))), nil
```

`1785416892000000000` interpreted as seconds is somewhere around the year
56,558,000. The query range becomes astronomically wide, so:

1. the scan is unbounded (39s on a dataset that answers in 0.37s), and
2. `FixPeriodPlanner` sizes a slice from that range and `makeslice` panics.

Prometheus clients legitimately send (fractional) seconds and Loki clients send
nanoseconds, so the unit has to be inferred from the magnitude of the value.

Note `reader/controller/tempo.go` already does magnitude detection for the Tempo
path (`epochToTime`), but only for seconds/milliseconds/nanoseconds ≥ 1e18, and
that helper is not used by the Loki/Prometheus path.

### Affected versions

Reproduced on `v4.3.1` and the parser is unchanged on `master` (v5), so v5 is
affected as well.

I have a patch with unit tests and will open a PR shortly.

ivarbh:
Fixes #885.

### Problem

`ParseTimeSecOrRFC` (`reader/controller/utils.go`) assumes every numeric time
parameter is expressed in **seconds**. Loki clients send **nanoseconds** —
Grafana Logs Drilldown (`grafana-lokiexplore-app`) calls
`/loki/api/v1/index/volume` with ns epochs:

```
GET /loki/api/v1/index/volume?query={host_name=~".+"}&start=1785416892000000000&end=1785417792000000000&limit=3
```

`1785416892000000000` read as seconds lands around the year 56,558,000, so the
query degenerates into a full table scan and `FixPeriodPlanner` sizes a slice
from that range:

```
panic: makeslice: len out of range
  reader/plannerV2.(*FixPeriodPlanner).Process.func2
```

Measured on a small dataset: the ns request returned 200 after **39.2s** and then
killed the process; the same instant at second precision returned in **0.37s**.

### Fix

Numeric epochs are disambiguated by magnitude — seconds, milli-, micro- or
nanoseconds. The thresholds are the epoch of ~1973-03-03 expressed in each unit,
so any realistic timestamp maps to exactly one unit and the ranges never overlap.

Two details worth calling out:

- **Integer epochs are parsed with `ParseInt`, not `ParseFloat`.** Nanosecond
  epochs (~1.8e18) exceed the 53-bit mantissa of a `float64`, so a float
  round-trip silently loses sub-microsecond precision. Fractional input
  (Prometheus sends fractional seconds) still goes through `ParseFloat` +
  `math.Modf`.
- **The Tempo path already had its own `epochToTime`** in
  `reader/controller/tempo.go`, but it only handled seconds/milliseconds/ns ≥ 1e18,
  so it misread microsecond epochs as milliseconds and ns epochs below 1e18 as
  milliseconds. Both paths now share one implementation, which fixes those cases too.

The compiled regexp is also hoisted to a package-level `var` instead of being
rebuilt on every request.

### Tests

`reader/controller/utils_test.go` (new) covers each unit, fractional seconds,
sub-second milli/micro precision, exact ns round-tripping, RFC3339, the empty
default, and the previously-broken Tempo inputs.

```
$ go test ./reader/controller/ -run 'TestParseTimeSecOrRFCMagnitudes|TestEpochToTimeSharedWithTempo' -v
--- PASS: TestParseTimeSecOrRFCMagnitudes (0.00s)
--- PASS: TestEpochToTimeSharedWithTempo (0.00s)
ok  	github.com/metrico/qryn/v5/reader/controller
```

`gofmt`, `go vet ./reader/controller/` and `go build ./...` are clean.

### Verification in a real deployment

Built from `v4.3.1` + this patch and deployed to our Kubernetes cluster
(4 Gigapipe instances, ClickHouse backend). The exact request that previously
panicked now returns `200` in `0.40s` with real data, and Grafana Logs Drilldown
works end to end — `index/volume`, `detected_labels`, `detected_fields`,
`patterns` and `query_range` all answer with ns-epoch parameters, no restarts.

### DIFF
--- a/reader/controller/tempo.go
+++ b/reader/controller/tempo.go
@@ -477,7 +477,7 @@ func parseTraceSearchParams(r *http.Request) (*traceSearchParams, error) {
 	if err != nil {
 		return nil, fmt.Errorf("start: %v", err)
 	}
-	res.Start = epochToTime(startS)
+	res.Start = epochToTime(startS, 0)
 	if startS == 0 {
 		res.Start = time.Now().Add(time.Hour * -6)
 	}
@@ -485,23 +485,11 @@ func parseTraceSearchParams(r *http.Request) (*traceSearchParams, error) {
 	if err != nil {
 		return nil, fmt.Errorf("end: %v", err)
 	}
-	res.End = epochToTime(endS)
+	res.End = epochToTime(endS, 0)
 	if endS == 0 {
 		res.End = time.Now()
 	}
 	return &res, nil
-}
-
-// epochToTime converts an epoch timestamp to time.Time, handling seconds, milliseconds, and nanoseconds.
-func epochToTime(v int64) time.Time {
-	switch {
-	case v >= 1e18:
-		return time.Unix(0, v)
-	case v >= 1e12:
-		return time.Unix(0, v*int64(time.Millisecond))
-	default:
-		return time.Unix(v, 0)
-	}
 }
 
 func orDefault(str string, def string) string {
--- a/reader/controller/utils.go
+++ b/reader/controller/utils.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"regexp"
 	"runtime/debug"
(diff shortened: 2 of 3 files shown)


rita7lopes (approved):
The PR looks good and adds the corresponding tests.
Thank you for fixing this bug, @ivarbh!

rita7lopes:
@ivarbh Please sign the CLA so we can merge the PR

ivarbh:
> @ivarbh Please sign the CLA so we can merge the PR

CLA signed

rita7lopes:
> > @ivarbh Please sign the CLA so we can merge the PR
> 
> CLA signed

Each PR needs its own signed CLA; I see you signed it for #887. This one must be signed too @ivarbh  https://github.com/metrico/gigapipe/pull/886#issuecomment-5163639045

ivarbh:
> > > @ivarbh Please sign the CLA so we can merge the PR
> > 
> > 
> > CLA signed
> 
> Each PR needs its own signed CLA; I see you signed it for #887. This one must be signed too @ivarbh [#886 (comment)](https://github.com/metrico/gigapipe/pull/886#issuecomment-5163639045)

Sorry my misunderstanding. I have signed the CLA for this PR also

PR #887, opened 2026-08-03
fix(logql): two planner panics behind Grafana Logs Drilldown 500s

ivarbh:
Both panics are hit by stock Grafana Logs Drilldown queries against a qryn/gigapipe Loki datasource, and both are recovered by `tamePanic` into an opaque HTTP 500.

### 1. Stale plan mode after `cancelJsonAndLogFmt`

`Plan()` computes `mode := decidePlanMode(script)` *before* `cancelJsonAndLogFmt(script)` mutates the pipeline list. When a chained `| json | logfmt` pair is cancelled, the stale mode can still be `planModeInternal`; `breakScript()` then returns a nil internal script and `internal/planner.Plan` feeds it to `shared.GetStrSelector`:

```
panic: runtime error: invalid memory address or nil pointer dereference
  shared.GetStrSelector      shared/funcs.go:63
  internal/planner.Plan      internal/planner/planner.go:14
```

Repro: `{service_name=~".+"} | json | logfmt | drop __error__`. Drilldown always emits `| json | logfmt`, so every "show logs" click 500s.

Fix: re-derive `mode` after the AST mutation.

### 2. `index out of range [-1]` in `AnalyzeMetrics15sShortcut`

`analyze.go:114` guards with `StrSel.Pipelines != nil`, but after `cancelJsonAndLogFmt` the slice is empty and non-nil, so the index computes `[-1]`:

```
panic: runtime error: index out of range [-1]
  clickhouse_planner.AnalyzeMetrics15sShortcut  analyze.go:115
```

Repro: `sum(count_over_time({service_name=~".+"} | json | logfmt [1m]))`. Drilldown's `LOG_STREAM_SELECTOR_EXPR` omits `| drop __error__`, so its histogram panel query trips this on load.

Fix: use the `len(...) > 0` idiom already used at `analyze.go:169` and `internal/planner/planner.go:165,184`.

### Tests

Two new tests, each red before its fix and green after:
- `reader/logql/logql_transpiler/planner_test.go` → `TestPlanCancelledJsonLogfmt`
- `reader/logql/logql_transpiler/clickhouse_planner/analyze_empty_pipeline_test.go` → `TestAnalyzeMetrics15sShortcutEmptyPipelines`

`go build ./...` clean, `go test ./reader/logql/...` all pass.

Related: #885 / #886 (nanosecond epoch parsing, the third Drilldown blocker).

### DIFF
--- a/reader/logql/logql_transpiler/clickhouse_planner/analyze.go
+++ b/reader/logql/logql_transpiler/clickhouse_planner/analyze.go
@@ -111,7 +111,7 @@ func AnalyzeMetrics15sShortcut(script *logql_parser.LogQLScript) bool {
 	if duration.Seconds() < 15 {
 		return false
 	}
-	if lraOrUnwrap.StrSel.Pipelines != nil &&
+	if len(lraOrUnwrap.StrSel.Pipelines) > 0 &&
 		lraOrUnwrap.StrSel.Pipelines[len(lraOrUnwrap.StrSel.Pipelines)-1].Unwrap != nil {
 		return false
 	}
--- a/reader/logql/logql_transpiler/clickhouse_planner/analyze_empty_pipeline_test.go
+++ b/reader/logql/logql_transpiler/clickhouse_planner/analyze_empty_pipeline_test.go
@@ -1,0 +1,32 @@
+package clickhouse_planner
+
+import (
+	"testing"
+
+	"github.com/metrico/qryn/v5/reader/logql/logql_parser"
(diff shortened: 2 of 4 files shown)


rita7lopes:
@ivarbh Thanks for fixing this bug!
Please sign the CLA so we can consider merging it

ivarbh:
recheck

rita7lopes (approved):
Thank you!

PR #899, opened 2026-08-11
Change partition Granularity to monthly

evan-zhao-nb:


lmangani:
@evan-zhao-nb could you kindly expand and provide some context/validtion/motivation on the proposed change and sign the CLA?

evan-zhao-nb:
@lmangani Hi, sorry I opened this PR into the upstream by accident when creating a PR into my own fork. 

If you're interested in hearing my motivation: 

My current use-case requires ingestion of logs from services that have been producing logs for quite a long time. 

The configuration of my log-forwarding daemon does not tail the log file and instead tries to send the entire file over (which sometimes results in >100d of logs). Due to the current CH table DDL, this results in a `Too many partitions for single INSERT block (>100)` error. Rather than changing the configuration in my CH cluster to support wider partitions for singular INSERTS, I found it easiest to just modify the DDL to partition by month instead, removing this error entirely. It's an opinionated take, but it seemed like the most architecturally sound way to deal with my current high-volume log ingestion situation.

If you'd like me to refine this PR more, it makes the most sense to introduce it as an ENV variable toggle so users can override the default behavior of partitioning by day and instead partition by month. It's also only relevant for logging telemetry rather than metrics, traces, or profiles since many logging daemons (e.g. Grafana Alloy) have a default behavior of frontloading the entire log file to the downstream telemetry store - unlike metrics/traces/profiles which are collected in realtime and have no prior history to send over upon initial instrumentation.

WDYT?

rita7lopes:
Hello @evan-zhao-nb 

Interesting case!
If you can submit a PR, as you said, introducing the TTL as an env var so users can override the default as per their use case, we would deeply appreciate it! 

Considering that samples_v3 is ordered by `timestamp_ns` it means that having monthly partitions instead of daily ones would cause no significant delays on recent queries like the last 30min and such. We would need the PR to be able to be sure (to run some tests).
Anyway, this matches clickhouse docs, so in general this would be a super nice PR if you still feel like doing it!

PR #906, opened 2026-08-18
promql: include the last step in the fill grid

janeblower:
## PromQL pushdown: cross-series aggregations drop series at the last step

`WITH FILL TO` is exclusive — it generates rows while the value is `< TO` — and
`range(start, end)` is `[start, end)`. Both fill paths were handed `ctx.To`
directly, so no evaluation row was ever created at the final step of the grid.
A series whose last real bucket fell earlier than that step had no row there,
and since consumers mask filled rows with `-If(..., source = 1)` and drop
`w_src = 0`, it disappeared from the step instead of being carried into it.

Cross-series aggregations therefore aggregated whatever subset happened to hold
a bucket in that exact slot. With 29 `up` series and a 60s scrape interval:

count(up) @ time=1787047200   -> 7    (step aligned)
count(up) @ time=1787047207   -> 29   (unaligned, fill reaches the previous node)

An instant query is a one-point window, so at a step-aligned timestamp that
subset *is* the whole answer — which is what shows up as dipping Grafana
Stat/Gauge panels, `stddev` collapsing to 0, and alert rules evaluating on a
fraction of their series. Range queries are wrong only at the final step.

The `arrayJoin` fallback for ClickHouse < 24.11 was worse: for a bucket landing
exactly on `ctx.To` the range collapsed to empty, dropping even the series that
did have a sample there — so the two paths were not producing the identical
rows they are documented to produce.

### Fix

Bound both fill paths at `ctx.To + 1`. `+1` rather than `+step` admits the node
at `ctx.To` and nothing past it.

### Testing

`go test ./reader/...` passes. Golden-SQL expectations updated, plus a test
asserting the bound is neither `ctx.To` (under) nor `ctx.To + step` (over) on
both paths.

Checked against ClickHouse 26.7.3 with three series — on
`ctx.To`, two earlier. At the last step, before: 1 series on the STALENESS path,
no rows at all on the arrayJoin path. After: 3 series on both, and the two paths
agree row for row.

Affects every v5.x (the pushdown became the default in v5.0.0).

Workaround for anyone hitting this on a released build: make the aggregation
argument an expression (`count(up * 1)`), which declines the pushdown and falls
back to the engine.


janeblower:
@rita7lopes 

Only e2e is red: 3 snapshots in qryn-test, each differing in exactly one point — the last step.

```
should read rate:  600  0.06666666666666667  ->  0.05
should sum + rate: 600  0.2                  ->  0.15000000000000002
exp: sum + rate:   600  0.2                  ->  0.15000000000000002
```

Test data is a counter stepping every 15s, last sample at t=585, `rate([1m])` at step 15s over 600s.

- **before:** no row at t=600, so the point came from `prolongSeriesIt` carrying the previous one forward - that's why the snapshot has `600 == 585 == 0.0667`
- **after:** the row exists and is computed like every other step: window `(540, 600]`, opening sample at 540, `(39-36)/60 = 0.05`. At t=585 the same formula gives `0.0667`, so the new point is consistent with the rest of the series.

We don't extrapolate a partially covered range the way Prometheus does, so `rate` now dips on the last step when the newest sample lags behind `t`. Mid-window that dip was always there; the carry-forward only hid it at the edge.

I'd rather just regenerate the three snapshots in qryn-test - the new values are what the formula gives at every other step, and the old ones were a carried duplicate. Matching Prometheus at the edge means extrapolating the covered interval, which changes `rate`/`increase`/`delta` across the whole range, not only the last step, so that belongs in its own PR. 
Want me to open the qryn-test PR, or will you regenerate them?

rita7lopes:
Hello @janeblower,

Thank you for your contribution!
Your fix exposed a problem we can now detect that was introduced earlier.
We will make a PR to solve it, so we can then merge yours.

Issue #917, opened 2026-08-21
Testing & Release Hardening Initiative


tsouza:
#### Why

Gigapipe's correctness surface has outgrown its test coverage. Today, PRs run a build, untagged unit tests (no race detector), one focused Content-Type regression against a live ClickHouse, and CodeQL — while releases are published by `build_release.yml` with **no test gate at all**. The transpilers (LogQL/PromQL/TraceQL → SQL), the hybrid PromQL execution path, the writer's ingestion pipeline, the schema migration scripts, and the recording-rules ruler all carry real regression risk that nothing currently pins.

This initiative builds a layered test harness inside this repository and restructures CI around a **two-tier fence**:

- **Merge gate** — a small, fixed set of fast required checks on every PR (build, race-enabled unit tests, lint, golden corpora, conformance, single-node schema migration). Target: under ~10 minutes.
- **Release gate** — the heavy lanes (parity vs reference Prometheus/Loki/Tempo, full roundtrip corpus walks, e2e stacks, chaos, perf). These trigger on every PR but short-circuit to a green no-op unless the event is a push to master, a nightly schedule, a manual dispatch, or a `release/*` PR. A **publish preflight** in `build_release.yml` refuses to release unless every release lane posted green on the exact commit being shipped.

Two principles run through everything:

1. **Ratchets over hard gates.** Where gigapipe deliberately diverges from reference behavior (15s range alignment, downsampled-table semantics), a hard-fail parity gate would be permanently red and get ignored. Instead, parity results are scored against a committed per-case baseline with a per-case reason classification; the gate fails only when a case *changes state*. Regressions are caught from day one; burndown of known divergences is visible and deliberate.
2. **Tests version with the code.** E2E lives in this repo under `test/e2e/` (compose stacks, Go drivers behind build tags, a deterministic all-signal seed generator). The external test repository will not be used; a behavior change and its test update land in the same PR.

A parallel workstream refactors the codebase for testability (constructor injection, one config package, a narrow DB-access seam, process-exit policy only in `cmd/`) — sequenced *behind* the black-box safety net so every refactor PR is provably behavior-preserving.

#### Milestones

Each milestone is a sub-issue of this one, in dependency order:

- [ ] #918 — M1 — Tooling groundwork + PromQL transpile-determinism fix
- [ ] #920 — M2 — Write→read roundtrip harness + regression bank (safety net complete)
- [ ] #922 — M3 — PromQL parity harness + release-gate skeleton + publish preflight
[...]

PR #925, opened 2026-08-21
M1: tooling groundwork + PromQL transpile-determinism fix

### LINKED ISSUE #918: M1 — Tooling groundwork + PromQL transpile-determinism fix
#### Goal

Establish the tooling floor every later milestone builds on, and land the one bug fix that's a hard prerequisite for anything that pins transpiler behavior later: `TranspileExpressionV2`'s pushdown-substitute naming used `rand.Int63()`, so the same query could emit different (but equivalent) SQL across runs.

#### Scope — groundwork

- [ ] Adopt `golangci-lint` with a committed config; wire it into CI.
- [ ] Add `-race` to the unit-test lane.
- [ ] Adopt the repo's Makefile as the single entry point for every test lane (`make build`, `make test-unit`, `make lint`, `make arch-lint`, `make fuzz HEAD=<head>`, …); CI invokes recipes, never raw commands.
- [ ] Replace `rand.Int63()` substitute naming in `reader/promql/promql_transpiler/optimizer/{vector_range,vector_agg}.go` with a deterministic per-transpile counter.
- [ ] **Determinism test**: transpile the same PromQL expression twice in `reader/promql/promql_transpiler`, assert the emitted SQL is byte-identical — direct regression coverage for the `rand.Int63()` → per-transpile-counter fix above.
- [ ] Add `go-arch-lint` in report-only mode to baseline current package coupling.
- [ ] **Parser pins**: table-driven tests per head pinning AST shape for representative queries (valid input parses to the expected root/field shapes the planners consume; invalid input fails with the documented error class) — cheap insurance against upstream parser-dependency bumps changing shapes underneath the transpilers.
- [ ] **Parser fuzz targets**: one native Go `FuzzParse` target per query head (LogQL, PromQL, TraceQL — the profiling head's selector parser joins in M5), seeded from representative queries, asserting parse-never-panics and parse-then-render round-trips where the grammar supports it. Wired as `make fuzz HEAD=<head> [DURATION=<duration>]` (default 60s). Local and manual-dispatch only — fuzzing is exploratory, never a CI gate; any crasher found is committed as a regression seed in the target's `testdata/` corpus so it replays forever as a plain test.

#### Acceptance

- Merge gate runs build + unit(-race) + lint + arch-lint(report) in under ~10 minutes.
- The determinism test is green and would catch a reintroduced `rand.Int63()`-style nondeterminism.
- Parser pins and `make fuzz HEAD=<head>` run clean per query head.

#### Non-goals

Any query-behavior pinning — decision-level or response-level, for any query head — deferred to M2 (response goldens) and M6 (decision pins). Profiling (M5 — needs its own fixture-format design). Any refactoring. Any ClickHouse-backed lane.

---
Originally scoped as groundwork + a golden-SQL corpus per query head (~130 fixtures across PromQL/LogQL/TraceQL). The corpus has been dropped: a static SQL-text contract bakes in today's emitted shape, including shapes that may themselves be bugs, so every fix becomes two edits (code + golden) instead of one. Query-behavior pinning now lands entirely in M2 (response-level goldens) and M6 (decision pins). See #917 for the full epic resequencing.

tsouza:
Closes #918

## Summary

Lands the M1 tooling floor: golangci-lint, `-race` on the unit-test lane, the
Makefile as the single entry point for build/test/lint/fuzz, go-arch-lint in
report-only mode — plus the one bug fix that's a hard prerequisite for anything
that pins transpiler behavior later.

- **Groundwork**: golangci-lint (scoped to what's actually idiomatic here — see
  the config's comments for why errcheck/structtag/composites are excluded),
  `-race` on the CI unit-test lane, `Makefile` targets (`build`/`test-unit`/
  `lint`/`arch-lint`/`fuzz`) as the single entry point for these lanes,
  go-arch-lint wired in report-only mode.
- **Determinism fix**: `TranspileExpressionV2`'s pushdown-substitute naming used
  `rand.Int63()`, so the same query could emit different (but equivalent) SQL
  across runs. Replaced with a per-transpile counter on `promql_parser.Expr`,
  now covered by a direct determinism test (transpile the same expression twice
  via the real engine path, assert byte-identical output).
- **Parser AST-shape pins** per query head (PromQL/LogQL/TraceQL): table-driven
  tests pinning that a representative valid query parses to the expected
  root/field shapes the planners consume, and that a representative invalid
  query fails with the documented error class — cheap insurance against
  upstream parser-dependency bumps changing shapes underneath the transpilers.
  Lives in its own `reader/internal/parserpin` package.
- **Parser fuzz targets**: one native Go `FuzzParse` target per query head,
  wired as `make fuzz HEAD=<head> [DURATION=<duration>]`. Local and
  manual-dispatch only — never a CI gate.

## What's not in this PR

The original scope included a golden-SQL corpus (153 fixtures pinning the
literal ClickHouse SQL text each PromQL/LogQL/TraceQL query lowers to). That's
been dropped: a static SQL-text contract bakes in today's emitted shape,
including shapes that may themselves be bugs, so every fix becomes two edits
(code + golden) instead of one. Query-behavior pinning now lands at two other
layers instead — decision pins for internal routing/pushdown choices, and
response-level goldens for the Loki/Prometheus/Tempo JSON contracts Grafana
actually depends on. Neither is part of this PR; both land in later milestones
of the testing epic (#917, resequenced accordingly — see the epic for the
updated milestone order).

## Test plan

- [x] `go build ./...`
- [x] `go test ./... -race -count=1` (whole module, clean)
- [x] `make lint` → 0 issues
- [x] `make arch-lint` → no warnings
- [x] `make fuzz HEAD=promql DURATION=2s` (and logql/traceql) → no crashes
- [x] New determinism test (`TestTranspileIsDeterministic`) run with
      `-race -count=2`, green
- [x] Manually swept the full diff for any reference to a prior/internal
      source project name and any AI-generated attribution text — none found


akvlad:
Hello @tsouza . 
This is a PR for unit testing. We thank you for that.
Although I don't see a lot of sense in this kind of unit tests, it can improve our codebase.
Please use a 3rd party test library like `go-snaps`, `autogold`, `gotest.tools/v3/golden` instead of the self-written `internal/goldentest` one.

Thanks.

tsouza:
Hi, thanks for the notes. 

> Please use a 3rd party test library like `go-snaps`, `autogold`, `gotest.tools/v3/golden` instead of the self-written `internal/goldentest` one.

The reason for writing it standalone is really to simplify things. I will evaluate those tools to see if any can be used without increasing complexity.

>Although I don't see a lot of sense in this kind of unit tests, it can improve our codebase.

This is part of a larger plan. Indeed this code change _alone_ does not makes sense, but since it's part of a much larger change, I decided to split. See https://github.com/metrico/gigapipe/issues/917 for more context

tsouza:
Rescoped this PR after further discussion. Dropped the golden-SQL corpus (153 SQL-text fixtures + the goldentest infra) — a static SQL-text contract pins today's emitted shape, including shapes that may themselves be bugs, so every fix meant two edits (code + golden) instead of one. Query-behavior pinning now lands at two other layers in later milestones instead: decision pins for internal routing/pushdown choices, and response-level goldens for the actual Loki/Prometheus/Tempo JSON contracts. See #917 for the updated milestone order.

What's left in this PR: tooling groundwork (golangci-lint, `-race`, Makefile entrypoints, go-arch-lint) and the `rand.Int63()` → per-transpile-counter determinism fix, now backed by a direct test instead of only being inferred from golden re-runs. Parser AST-shape pins and fuzz targets are unaffected.

tsouza:
Pushed a follow-up commit enriching the parser AST-shape pins and fuzz seed corpus for all three heads (they were thin: 14-17 pins / 10-14 seeds each). New material is validated against this repo's actual parser (not assumed) and drawn from permissively-licensed upstream sources already vendored for reference elsewhere — prometheus/compliance's query set for PromQL, grafana/loki's pkg/logql/bench registry for LogQL. Several new pins are negative cases documenting real current parser gaps (unwrap conversion-function syntax, empty by()/without() on range aggregations, TraceQL's bare ">"/"<"/"!(" forms) rather than assuming upstream grammar support carries over.

rita7lopes:
Thanks for the rescope — dropping the SQL-text corpus was the right call, and the PR is much closer to landable.

Three more things needed before merging the PR.

### 1. Remove the parser AST-shape pins

```
reader/promql/promql_parser/parserpin_test.go      (425)
reader/traceql/traceql_parser/parserpin_test.go    (320)
reader/logql/logql_parser/parserpin_test.go        (271)
reader/internal/parserpin/parserpin.go             (70)
```

The four come out cleanly as a unit — the fuzz targets don't import the helper, and nothing else consumes it.

The reasoning is the one that retired the SQL corpus, one layer up. These pin the *shape* of a parsed AST rather than a behaviour, and the coverage already exists at the behaviour level: `reader/promql/promql_transpiler/transpiler_test.go:17-22` parses and transpiles `sum by (a) (rate(http_requests_total{...}[5m]))` and asserts on the emitted result, so an upstream AST change breaks that first and with a clearer message. There are ~2,370 lines of transpiler tests on master already doing this.

Two categories are worth calling out specifically:

- The 37 LogQL/TraceQL `Check` pins assert our *own* participle struct shapes (`StrSelCmds`, `Pipelines[0].LineFilter.Fn`, `ByOrWithoutSuffix`). There is no upstream parser behind those two heads, so they don't protect against a dependency bump — they fire when we refactor our own AST.
- Five of the `WantErr` pins assert that we reject queries which are valid in the real languages, which pins the gap open: a contributor who implements the feature gets a red test telling them not to.

That last point is the most valuable thing in this PR. Those five are real compatibility bugs, now filed with the upstream grammar citations: #986, #987, #988, #989, #990. Checking them turned up more than the pins covered — 11 of TraceQL's 14 structural operators are unsupported, and all three LogQL `unwrap` conversion functions, not just `duration()`. Without this work none of that would have surfaced.

### 2. Match the Go toolchain to go.mod

The arch-lint job pins `go-version: '1.26'`, but master is on 1.27.1 (#969) and #933 moved CI to `go-version-file: go.mod`.

### 3. Confirm the golangci-lint pin still builds

v2.13.1 was pinned when go.mod was on an older Go; worth re-checking against 1.27.1.

---

**Keeping:** the determinism fix and its direct test — "same input produces the same output" is exactly the kind of invariant that earns a permanent test — plus the three `FuzzParse` targets, golangci-lint, `-race`, go-arch-lint, and the Makefile entrypoints.

Context for the wider plan is in #917 and the per-milestone comments.

---

Sorry for the delay in replying. Between the work on #972 and the other PRs from @janeblower, we needed to agree on the path forward first so the tests would match the plan rather than the other way round.

Let us know whether you'd like to make the three changes, or if you'd rather I do it.

rita7lopes (changes requested):
Check my comment above

PR #935, opened 2026-08-24
Deps/influx line protocol

janeblower:
First of a possible series - I've been going through what our dependency graph
actually contains, and this looked like the best place to start.

The influx write path used telegraf's line-protocol stream parser. That single
import pulls in telegraf's config machinery - 68 config-migration packages, the
plugin registries, memguard, cel-go, compose-go - and all of it gets linked into
the binary.

Replaced it with `influxdata/line-protocol/v2`, the library telegraf parses with
internally: no dependencies of its own, already in our module graph. Timestamp
handling is unchanged - the value is multiplied by the requested precision, and a
line without one gets the arrival time truncated to that precision.

Two parsers differ in one place, so it's handled explicitly: telegraf built tags
through `metric.AddTag`, which overwrote an existing key, while decoding straight
off the wire would append both. A line like `cpu,host=a,host=b` would then have
split into a second series. The decoder now deduplicates tag keys and keeps the
last value, so such input parses exactly as before.

The influx path had no tests, so this adds them for metrics, logs, missing
timestamps, duplicate tags and malformed lines.

881 > 623 modules, 1396 > 1243 packages, 68.9 > 60.4 MB binary. `go test ./...`
is green.


akvlad:
Hello @janeblower .
Thank you for a great improvement of the writer functionality.
Could you please describe the configuration of any standard influx client, like otel collector or anything else to send some data into the system, so we can test the changes manually?

Great PR. Thanks.

janeblower:
@akvlad Thanks!

Attached is a small stand for testing this by hand: `docker-compose.yml`, the client
configs, and a README with the commands.

It brings up one ClickHouse and two gigapipe instances against separate databases —
this branch on :3100 and `ghcr.io/metrico/gigapipe:latest` on :3200 — so the same
payload can be sent to both and the results diffed. Telegraf and the OTel collector
sit behind a `clients` profile.

```sh
docker compose up -d --build
curl -XPOST 'http://localhost:3100/influx/api/v2/write?precision=ns' \
  --data-binary "cpu,host=alpha usage_idle=91.2 $(( $(date +%s) * 1000000000 ))"
```

Two things the README spells out, since the clients disagree here: Telegraf wants
http://gigapipe:3100/influx and appends /api/v2/write itself, while the OTel
collector's influxdb exporter needs the full path. Neither token is checked.

Running the comparison myself, the two versions store byte-identical rows for
escaped names, duplicate tag keys, string/int/bool fields and missing timestamps;
only the wording of parse errors differs.

[stand-influx.zip](https://github.com/user-attachments/files/31530354/stand-influx.zip)

rita7lopes (approved):
Very nice PR @janeblower Thank you very much!!
There's a change in behaviour, a change for the best:
Before if a misconfigured client sent:
`body: cpu,host=a value=1 1600000000000000000   precision=s` this is a precision mismatch: client sends ns timestamps, URL says s or ms. 
Code in master: accepted, stored ts = -3710740494427357184
With this PR: "invalid timestamp: line-protocol value out of range", batch aborted

Storing and acceting garbage was wrong, this is a better outcome.

Please sort the conflicts with a rebase onto master, or let me know if we should do it. Otherwise, the PR is ready to be merged

janeblower:
@rita7lopes rebase done

PR #940, opened 2026-08-25
refactor(writer): extract a transport-agnostic ingest core with signal-scoped service resolvers

44YHC:
Base of a 3-PR stack (this → OTLP metrics → OTLP/gRPC receiver). Squashed final state:

- `IngestParsed(ctx, parser, svcs)` carries the parse → push pipeline that `doParse` previously inlined, so ingest paths that do not start from an `http.Request` can reuse it; `doParse` now just resolves its services from the request context and delegates.
- `InsertServices` bundles the per-tenant insert services, populated by signal-scoped resolvers (`ResolveLogServices`, `ResolveTraceServices`, `ResolveProfileServices`, `ResolveMetricServices`): each signal resolves only its own services.
- `Node` — the fingerprint-cache namespace (`FPCache.DB(Node)`) — comes from the **time-series** service. The cache gates exactly one thing: whether a series' `time_series` row is appended, and those rows are pushed to `Ts`; with an empty tenant DSN the static registry picks an independent random node per getter call, so any other choice can mark a fingerprint seen on a node that never received the row — silent series loss until the cache TTL resets. `TestResolveLogServicesNodeFollowsTimeSeries` pins this.
- The HTTP middlewares resolve through the same helpers, keeping every ingest path's service resolution in one place.

### DIFF
--- a/writer/controller/builder.go
+++ b/writer/controller/builder.go
@@ -198,19 +198,13 @@ func doLogsPattern(s *model.TimeSamplesData) {
 	controller.ClusterLines(s.MMessage, s.MFingerprint, s.MTimestampNS)
 }
 
-func doParse(r *http.Request, parser Parser) error {
-	reader := getBodyStream(r)
-	tsService := getService(r, utils.ContextKeyTsService)
-	splService := getService(r, utils.ContextKeySplService)
-	spanAttrsService := getService(r, utils.ContextKeySpanAttrsService)
-	spansService := getService(r, utils.ContextKeySpansService)
-	profileService := getService(r, utils.ContextKeyProfileService)
-	node := r.Context().Value(utils.ContextKeyNode).(string)
-
-	//var promises []chan error
+// IngestParsed runs parser and pushes each ParserResponse to the given
+// per-tenant insert services. It is the transport-agnostic core shared by the
+// HTTP handlers and the gRPC receiver.
+func IngestParsed(ctx context.Context, parser Parser, svcs InsertServices) error {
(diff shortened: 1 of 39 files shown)


writer/controller/builder.go
```diff
@@ -198,19 +198,13 @@ func doLogsPattern(s *model.TimeSamplesData) {
...
-	//var promises []chan error
+// IngestParsed runs parser and pushes each ParserResponse to the given
+// per-tenant insert services. It is the transport-agnostic core shared by the
+// HTTP handlers and the gRPC receiver.
+func IngestParsed(ctx context.Context, parser Parser, svcs InsertServices) error {
+	fpNode := FPCache.DB(svcs.Node)
 	var promises []*promise.Promise[uint32]
-	var err error = nil
-	res := parser(r.Context(), reader, FPCache.DB(node))
+	res := parser(ctx, nil, fpNode)
```
rita7lopes: IngestParsed accepts the exported Parser type, whose signature includes an io.Reader it never supplies. Both of these satisfy Parser and both compile:

the pre-bound closure above: ignores the nil, uses its captured reader
a raw parser like unmarshal.UnmarshallMetricsWriteProtoV2: reads the nil
The second one hits io.ReadAll(ctx.bodyReader) in withBufferedBody (writer/utils/unmarshal/builder.go:492) with a nil interface: runtime error: invalid memory address or nil pointer dereference.

rita7lopes (changes requested):

rita7lopes (approved):

PR #941, opened 2026-08-25
feat(writer): OTLP metrics ingestion on /v1/metrics

44YHC:
Middle of the 3-PR stack — **base is `refactor/writer-ingest-core` (#940)**; retarget to `master` after #940 merges. Squashed final state:

- `/v1/metrics` accepts OTLP/HTTP exports in binary protobuf and JSON, responding per the OTLP/HTTP spec: 200 with `partial_success` when data points were rejected, 400 `google.rpc.Status` for undecodable payloads, 413 for oversize bodies, 503 for transient ingest failures.
- The decode layer translates OTLP metrics to Prometheus-shaped series per the OTel↔Prometheus compatibility spec: name/unit/suffix mapping, `job`/`instance` from `service.*`, `target_info` (one sample per resource per export at the latest accepted data-point timestamp), `otel_scope_*` labels with identity-label collisions dropped, colliding sanitized attribute keys concatenated with `;` in original-key order, exact exponential-histogram bucket bounds, delta temporality rejected.
- The payload bound is `system_settings.otlp_max_message_size` from cloki-config v0.0.96 (env `QRYN_SYSTEM_SETTINGS_OTLP_MAX_MESSAGE_SIZE`), default 64 MiB, applied to the decompressed body.
- `withPreParsedBody` lets a parser run over an already-decoded proto object, which is how the handler feeds the shared ingest core (and how the gRPC receiver PR feeds it next).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@
 - [Configuration & Docker](docs/configuration.md) — environment variables, Docker quickstart, cross-cluster setup
 - [Profiling API](docs/profiling-api.md) — profiling endpoints, DOT format export, Graphviz visualization
 - [OTLP Profiles Ingestion](docs/otlp-profiles.md) — OpenTelemetry profiles signal ingestion via `/v1development/profiles`
+- [OTLP Metrics Ingestion](docs/otlp-metrics.md) — OpenTelemetry metrics signal ingestion via `/v1/metrics` (protobuf + JSON)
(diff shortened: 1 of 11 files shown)


rita7lopes (approved):
There's a lot of parts talking about gRPC and referring to docs this PR so we need to make sure these features land together or we should remove the references before merging to main

PR #942, opened 2026-08-25
feat(writer): OTLP/gRPC receiver for traces, logs, metrics, and profiles

44YHC:
Top of the 3-PR stack — **base is `feat/otlp-http-metrics` (#941)**; retarget as the stack merges. Squashed final state:

- All four OTLP collector services are served over gRPC on the writer's existing HTTP port: `writergrpc.Mux` dispatches HTTP/2 requests whose Content-Type starts with `application/grpc` to a grpc-go server and everything else to the regular router, with `http.Protocols` enabling cleartext (prior-knowledge) HTTP/2. Reader-only nodes skip the wrapper.
- Handlers decode nothing themselves: grpc-go decodes the request, and `OTLP{Traces,Logs,Metrics}FromData` / `OTLPProfilesFromProfiles` feed the already-decoded objects to the shared ingest core through `withPreParsedBody`. Tenant metadata (DSN, TTL, async mode) is resolved from gRPC metadata the same way the HTTP middlewares resolve headers.
- Logging, panic recovery, and basic auth are ported as unary **and** stream interceptor chains (streaming RPCs never hit unary interceptors); ingest failures return `codes.Unavailable` — retryable per the OTLP spec, unlike `codes.Internal` — with an opaque message so backend DSNs never reach clients.
- The receive limit is the shared OTLP payload bound (`system_settings.otlp_max_message_size`), so a batch accepted over OTLP/HTTP is never rejected over gRPC; grpc-go's own default is 4 MiB.
- Because gRPC rides the shared `http.Server`, `http.Server.Shutdown` is what drains in-flight RPCs; `grpc.Server.GracefulStop` does not apply to ServeHTTP-served connections.

### DIFF
--- a/cmd/gigapipe/main.go
+++ b/cmd/gigapipe/main.go
@@ -381,6 +381,35 @@ func stop() {
 // background flushes to complete before forcibly exiting.
 const shutdownTimeout = 30 * time.Second
 
+// servesGRPC reports whether a node in this mode mounts the OTLP/gRPC
+// receiver. The receiver ingests through the write path (controller.Registry,
+// populated by writer.Init), so it follows exactly the modes that boot the
+// writer subsystem — see bootSequence, whose in("all", "writer", "") gate this
+// mirrors. TestServesGRPCFollowsWriterModes pins the two together.
+func servesGRPC(mode string) bool {
+	return mode == "all" || mode == "writer" || mode == ""
+}
+
+// httpRoot returns the root handler and protocol set for the shared HTTP
+// server. It is split out of httpStart so the mode gating can be asserted
+// directly: httpStart binds a listener and blocks on signals, so the decision
+// is untestable while it stays inline.
+//
+// Both return values are gated on the same condition, deliberately. Reader-only
+// nodes have no write path, so they get neither the OTLP/gRPC dispatcher nor
+// the protocol set that carries it: writergrpc.Protocols() enables cleartext
+// (prior-knowledge) HTTP/2, which exists only to carry gRPC. Setting it on a
+// node with no gRPC handler behind it would widen the protocols that node
+// accepts for no gain. A nil *http.Protocols leaves net/http's default
+// (HTTP/1 plus HTTP/2 over TLS), which is what reader-only nodes served before
+// the receiver existed.
+func httpRoot(server http.Handler, mode string, grpcOpts writergrpc.Options) (http.Handler, *http.Protocols) {
+	if !servesGRPC(mode) {
+		return server, nil
+	}
+	return writergrpc.Mux(server, grpcOpts), writergrpc.Protocols()
+}
+
 func httpStart(server *mux.Router, httpURL string, mode string, grpcOpts writergrpc.Options) {
 	logger.Info("Starting service")
 	var err error
@@ -390,13 +419,8 @@ func httpStart(server *mux.Router, httpURL string, mode string, grpcOpts writerg
 		panic(err)
 	}
 
-	// Reader-only nodes have no write path (controller.Registry is nil there),
-	// so only wrap with the OTLP/gRPC dispatcher when this node writes.
-	var root http.Handler = server
-	if mode == "all" || mode == "writer" || mode == "" {
-		root = writergrpc.Mux(server, grpcOpts)
-	}
-	httpServer := &http.Server{Handler: root, Protocols: writergrpc.Protocols()}
+	root, protocols := httpRoot(server, mode, grpcOpts)
+	httpServer := &http.Server{Handler: root, Protocols: protocols}
 
 	// Start serving in a goroutine so we can block on the signal below.
 	go func() {
--- a/cmd/gigapipe/main.go
+++ b/cmd/gigapipe/main.go
@@ -328,7 +328,11 @@ func start() {
 		step.run(cfg, app)
 	}
 	httpURL := fmt.Sprintf("%s:%d", cfg.Setting.HTTP_SETTINGS.Host, cfg.Setting.HTTP_SETTINGS.Port)
-	httpStart(app, httpURL, cfg.Setting.SYSTEM_SETTINGS.Mode)
+	grpcOpts := writergrpc.Options{
+		BasicAuthUser: cfg.Setting.AUTH_SETTINGS.BASIC.Username,
+		BasicAuthPass: cfg.Setting.AUTH_SETTINGS.BASIC.Password,
+	}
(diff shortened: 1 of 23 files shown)


cmd/gigapipe/main.go
```diff
@@ -377,7 +381,36 @@ func stop() {
...
 
-func httpStart(server *mux.Router, httpURL string, mode string) {
+// servesGRPC reports whether a node in this mode mounts the OTLP/gRPC
+// receiver. The receiver ingests through the write path (controller.Registry,
+// populated by writer.Init), so it follows exactly the modes that boot the
+// writer subsystem — see bootSequence, whose in("all", "writer", "") gate this
+// mirrors. TestServesGRPCFollowsWriterModes pins the two together.
+func servesGRPC(mode string) bool {
+	return mode == "all" || mode == "writer" || mode == ""
+}
```
44YHC, line 391: `servesGRPC` restates `bootSequence`'s `in("all", "writer", "")` literal, and `TestServesGRPCFollowsWriterModes` keeps the copy honest — but only across the five modes hardcoded in the test's slice. Add a sixth writer-booting mode to `bootSequence` and both the gate and the test miss it silently.

Deriving it makes the invariant structural instead of test-enforced:

```suggestion
// servesGRPC reports whether a node in this mode mounts the OTLP/gRPC
// receiver. The receiver ingests through controller.Registry, which only
// writer.Init populates, so it follows the modes that boot the writer.
func servesGRPC(mode string) bool {
	return slices.ContainsFunc(bootSequence(mode), func(s bootStep) bool { return s.name == "writer" })
}
```

I ran this in a worktree: it compiles (`slices` is already imported) and both `httpRoot` tests pass unmodified. It also makes `TestServesGRPCFollowsWriterModes` redundant — the drift it guards against becomes unrepresentable. `bootSequence` is called once at startup, so the allocation is irrelevant.

cmd/gigapipe/main.go
```diff
@@ -377,7 +381,36 @@ func stop() {
...
+// is untestable while it stays inline.
+//
+// Both return values are gated on the same condition, deliberately. Reader-only
+// nodes have no write path, so they get neither the OTLP/gRPC dispatcher nor
+// the protocol set that carries it: writergrpc.Protocols() enables cleartext
+// (prior-knowledge) HTTP/2, which exists only to carry gRPC. Setting it on a
+// node with no gRPC handler behind it would widen the protocols that node
+// accepts for no gain. A nil *http.Protocols leaves net/http's default
+// (HTTP/1 plus HTTP/2 over TLS), which is what reader-only nodes served before
+// the receiver existed.
```
44YHC: This doc is 13 lines for a 6-line function — the commit adds ~26 comment lines for 12 lines of code in this file — and it reads as close to a verbatim paste of the commit message. Three parts I'd cut:

- L404-405, *"which is what reader-only nodes served before the receiver existed"* — references old behavior. A future reader with no memory of this PR can't verify it and doesn't need it. Repeated again at `main_test.go:180-181`.
- L401-403, *"Setting it on a node with no gRPC handler behind it would widen the protocols..."* — narrates the defect being fixed. That's commit-message material, and it's already there.
- L394-396, *"It is split out of httpStart so the mode gating can be asserted directly..."* — refactor rationale, also already in the commit message.

The rule the code follows is one sentence: the gRPC dispatcher and the h2c protocol set are a single decision, because h2c exists only to carry gRPC. The rest is history the commit message already carries.

cmd/gigapipe/main_test.go
```diff
@@ -119,3 +122,70 @@ func TestBootSequenceModesScope(t *testing.T) {
...
+// TestHTTPRootWriterModes asserts that a node which serves gRPC gets both
+// halves: the dispatcher wrapping the router, and the protocol set that
+// carries it. Cleartext HTTP/2 must be enabled, since gRPC rides
+// prior-knowledge h2c on this port.
+func TestHTTPRootWriterModes(t *testing.T) {
+	for _, mode := range []string{"all", "writer", ""} {
+		t.Run("mode="+mode, func(t *testing.T) {
+			base := &nopHandler{}
+			root, protocols := httpRoot(base, mode, writergrpc.Options{})
+			if root == http.Handler(base) {
```
44YHC: Minor: this proves *something* wrapped the router, not that it's the gRPC dispatcher — any wrapper passes. Fine at this level, just weaker than the failure message reads.

If you want it to bite, assert on behavior instead: push an HTTP/2 request with `Content-Type: application/grpc` through `root` and check it never reaches `base`.

44YHC (commented):
Reviewed the head commit (`f0540a0`) — fetched it, read the surrounding boot/serve path, and ran `go vet` plus `go test ./cmd/gigapipe/` locally. Vet is clean and all tests pass.

**The change is correct and I found no bugs.** Two gates that had to agree were separate; folding them into one function that returns both values is the right shape.

One framing note for other reviewers: nothing was actually broken before this. `httpServer.Serve(listener)` is cleartext-only (no `ServeTLS` anywhere in the binary), and an h2c client hitting a reader would have been dispatched to the normal router by `Mux`'s content-type check anyway — it just would have arrived over HTTP/2. So this is hardening/consistency rather than a defect fix. By the same token, `nil` Protocols costs readers nothing: net/http's default is HTTP/1 plus HTTP/2 *over TLS*, and TLS never happens on this listener. `docs/otlp-grpc.md` already documents reader-only nodes as not mounting the receiver, so the docs stay accurate.

The main suggestion is inline: derive the gate from `bootSequence` instead of restating its mode literal. I verified that version compiles and that both new tests pass against it unmodified.

Minor:
- The commit is scoped `fix(writer):` but touches only `cmd/gigapipe/`.
- `TestHTTPRootWriterModes` is added but not mentioned in the commit message, which names only the other two tests.

cmd/gigapipe/main.go
rita7lopes: applied in https://github.com/metrico/gigapipe/commit/98988d19a479a16a426a0323e5b3850137eaa541

cmd/gigapipe/main.go
rita7lopes: applied in https://github.com/metrico/gigapipe/commit/98988d19a479a16a426a0323e5b3850137eaa541

cmd/gigapipe/main_test.go
rita7lopes: applied in https://github.com/metrico/gigapipe/commit/98988d19a479a16a426a0323e5b3850137eaa541

PR #946, opened 2026-08-28
reader: hold the PromQL engine directly instead of api_v1.API

janeblower:
Another one in the dependency cleanup series, after #935.

The query_range router built a `prometheus/web/api/v1` API struct only to park the PromQL engine in it. The controllers read back exactly one field, `QueryEngine`. The rest stayed nil, and `CORSOrigin` was dead too — the API value is never registered as a handler, and CORS is served by our own middleware.

So the field is now just the engine. Four lines of code, the rest is `go mod tidy`.

That import pulled in `prometheus/config`, `rules` and `notifier`, and with them:

- the Azure SDK with MSAL, and twelve aws-sdk-go-v2 modules
- `alertmanager v0.33.1` with all twenty `go-openapi` packages
- `google.golang.org/api`, `cloud.google.com/go/auth`, `genproto`
- an older generation of otel-collector modules (v1.63.0 / v0.157.0)

|          | before  | after   |
|----------|---------|---------|
| modules  | 213     | 124     |
| packages | 1237    | 822     |
| binary   | 61.6 MB | 53.8 MB |

No behaviour change on `/api/v1/query` and `/api/v1/query_range`. `go build`, `go vet` and `go test ./...` unchanged.


janeblower:
@rita7lopes can you check it? :)

rita7lopes (approved):
Hey @janeblower,

Thank you! PR looks good and makes sense!
I think it just need a rebase onto main. Let me know if you want us to do it or if you do it yourself.
After, it's ready to be merged

janeblower:
@rita7lopes 
I think it would be better if you do the rebase yourself before merging.

PR #952, opened 2026-08-29
fix(promql): emit stale markers to stop lookback stacking (#931)

### LINKED ISSUE #931: Unexpected Gigapipe PromQL Lookback / Staleness Behavior
One could argue this is not a bug / working as designed, but is a bit unexpected coming from regular Prometheus

### Summary

Gigapipe's effective lookback window for instant vector selectors in range queries can be **up to ~10 minutes**, roughly double the standard Prometheus 5-minute staleness delta. This is caused by two independent 5-minute windows stacking on top of each other.

#### 1. SQL Layer: `const staleness = 5 * time.Minute`

**File:** `reader/promql/promql_transpiler/planner/shared.go`

The ClickHouse-accelerated query planner uses a hardcoded 5-minute staleness constant. At each evaluation step, the SQL query looks back up to 5 minutes for the most recent sample. This mirrors Prometheus's default `--query.lookback-delta`.

This value is a compile-time constant and is **not configurable** at runtime.

#### 2. Iterator Layer: `prolongSeriesIt` (forward-fill up to 5 minutes)

**File:** `reader/model/prometheus.go`

When the Prometheus engine calls `Select()` for an instant vector selector or rate function during a range query, gigapipe returns a `prolongSeriesIt` iterator. This iterator **forward-fills** the last real sample at every step until either:
- A new real sample arrives, or
- The current timestamp exceeds `prev.TimestampMs + 300000` (300,000 ms = 5 minutes)

**Condition for activation** (`reader/service/prom_queryable.go:204`):
```go
func (c *CLokiQuerier) isProlong(hints *storage.SelectHints, matchers []*labels.Matcher) bool {
    // disabled for substituted (accelerated) expressions
    return (hints.Func is "rate"/"deriv"/"delta" or "") && hints.Step != 0
}
```

3. Prometheus Engine LookbackDelta: default 5 minutes

reader/router/prometheus_query_range.go

The Prometheus engine is initialized with LookbackDelta: 0, which means "use the default" — 5 minutes. The engine applies its own lookback on top of whatever samples the iterator returns.

Combined Effect

Real sample at T
  → prolongSeriesIt fills forward from T to T+5m (step by step)
  → Prometheus engine can look back 5m from any evaluation step to find a filled sample
  → A sample at T can influence steps as late as T + ~10m

In standard Prometheus, a sample at T is only visible until T+5m. In gigapipe, it can remain visible until approximately T+10m.

Reproducibility

This affects any range query (/api/v1/query_range) with:

- An instant vector selector (e.g., metric_name{label="value"})
- Rate/delta functions (e.g., rate(metric[5m]))
- A non-zero step

None of these values are configurable

mgai-rchung:
## Summary

Fixes #931 — the effective PromQL lookback for instant vector selectors and `rate`/`deriv`/`delta` in range queries could be up to ~10 minutes, roughly double Prometheus's standard 5-minute staleness delta, because two independent 5-minute windows stacked:

1. the iterator forward-filled the last value on the step grid for 5m, and
2. the PromQL engine then applied its own 5m `LookbackDelta` on top of that.

## Approach

Emit a single Prometheus stale marker (`value.StaleNaN`) at the lookback boundary instead of forward-filling. The engine carries the last real value forward across its own 5m lookback and the marker terminates it exactly there, so total visibility is ~5m — the same mechanism Prometheus uses when a series disappears, and never the stacked ~10m.

A run ends only when the next real sample is more than the lookback delta (5m) away, so genuinely **sparse** series whose samples are within 5m are still bridged by the engine's lookback, exactly as in upstream Prometheus. When a stale series later **reappears**, the marker sits in the gap and the new sample simply starts a fresh run.

> Note on push vs pull: Prometheus's ingest-time "series went away" detection is a pull/scrape capability. For pushed data (OTLP, remote-write) gigapipe cannot detect disappearance at ingest, so this synthesizes the equivalent staleness at query time — reproducing the observable 5m-then-stale behavior without ingest-time detection.

## Changes

Both code paths are covered:

- **`prolongSeriesIt`** (non-substitute instant/rate selectors), `reader/model/prometheus.go`: yields the real samples untouched and emits one marker at `lastSample + LookbackDeltaMs` when a run ends (gap > 5m, or series end). No more step-grid fill.
- **Substitute-backed (SQL-filled) path**, `reader/service/prom_queryable.go`: gated on `isSQLFilled(matchers)` (a registered substitute for the `__name__`), which is the exact set whose rows `FillGapsPlanner` already forward-filled 5m. `appendStaleMarker()` caps that by adding one marker one step past the last returned sample in `Select()`, unless the series is still live at the query-window edge. Note: this is **not** `!isProlong` — non-substitute instant-vector functions (`abs`, `topk`, `histogram_quantile`, ...) are also `Prolong=false` but are planned by `HintsPlanner`'s `argMax` regroup (no fill), so they must not be marked.

`StaleMarkerValue` and `LookbackDeltaMs` are defined once in `reader/model` as the single source of truth.

## Testing

- Iterator unit tests (`reader/model`) including sparse-gap bridging and stale-then-reappear.
- `appendStaleMarker` unit tests (`reader/service`).
- Engine/iterator integration test (`reader/router`) that asserts **both** ~5m carry-forward within the lookback window **and** no ~10m stacking. Verified it fails against the old forward-fill behavior and passes with the fix.

All `go build ./...` and `go test ./reader/...` pass.

### DIFF
--- a/reader/service/prom_queryable.go
+++ b/reader/service/prom_queryable.go
@@ -210,6 +210,39 @@ func (c *CLokiQuerier) isProlong(hints *storage.SelectHints, matchers []*labels.
 	return (slices.Contains(rateFunctions, hints.Func) || hints.Func == "") && hints.Step != 0
 }
 
+// appendStaleMarker terminates an accelerated (substitute-backed) series with a
+// Prometheus stale marker one step past the last sample the SQL layer returned.
+//
+// The accelerated path builds a series with Prolong=false directly from
+// ClickHouse rows, and the SQL densifier (planner.fillGaps) already forward-fills
+// each real bucket for 5 minutes (planner.staleness) on the step grid. So the
+// last returned sample already sits at lastReal + 5m: the 5m carry is baked into
+// the rows. The problem in issue #931 is that the engine then applies its OWN 5m
+// LookbackDelta on top of that, stacking to ~10m. Appending a stale marker one
+// step past the last returned sample caps the already-filled data at the 5m
+// boundary so the engine cannot extend it any further.
+//
+// (This differs from the raw iterator path in reader/model, which does not
+// SQL-fill and therefore places its marker at lastReal + LookbackDeltaMs to
+// provide the 5m carry itself. Both paths yield ~5m total visibility.)
+//
+// The marker is only appended when the series genuinely stopped before the end
+// of the query window: if the last sample is within one step of queryEndMs the
+// series is still live and Prometheus would not stale-mark it. A zero/unknown
+// step is treated as "cannot place a marker" and is left untouched.
+func appendStaleMarker(samples []model.Sample, stepMs int64, queryEndMs int64) []model.Sample {
+	if len(samples) == 0 || stepMs <= 0 {
+		return samples
+	}
+	markerTs := samples[len(samples)-1].TimestampMs + stepMs
+	if markerTs > queryEndMs {
+		// The series runs up to (or past) the query edge; it did not stop, so
+		// no stale marker - the query window itself truncates it.
+		return samples
+	}
+	return append(samples, model.Sample{TimestampMs: markerTs, Value: model.StaleMarkerValue})
+}
+
 func (c *CLokiQuerier) Select(ctx context.Context, sortSeries bool, hints *storage.SelectHints,
 	matchers ...*labels.Matcher) storage.SeriesSet {
 
@@ -290,6 +323,10 @@ func (c *CLokiQuerier) Select(ctx context.Context, sortSeries bool, hints *stora
 			if len(res.Series) > 0 && q.MapResult != nil {
 				res.Series[len(res.Series)-1].Samples = q.MapResult(res.Series[len(res.Series)-1].Samples)
 			}
+			if len(res.Series) > 0 && !isProlong {
+				res.Series[len(res.Series)-1].Samples = appendStaleMarker(
+					res.Series[len(res.Series)-1].Samples, hints.Step, hints.End)
+			}
 			res.Series = append(res.Series, &model.SeriesV2{
 				LabelsGetter: lblsGetter,
 				Fp:           fp,
@@ -305,6 +342,10 @@ func (c *CLokiQuerier) Select(ctx context.Context, sortSeries bool, hints *stora
 	}
 	if len(res.Series) > 0 && q.MapResult != nil {
 		res.Series[len(res.Series)-1].Samples = q.MapResult(res.Series[len(res.Series)-1].Samples)
+	}
+	if len(res.Series) > 0 && !isProlong {
+		res.Series[len(res.Series)-1].Samples = appendStaleMarker(
+			res.Series[len(res.Series)-1].Samples, hints.Step, hints.End)
 	}
 	err = lblsGetter.Fetch()
 	if err != nil {
--- a/reader/service/prom_queryable.go
+++ b/reader/service/prom_queryable.go
@@ -210,28 +210,40 @@ func (c *CLokiQuerier) isProlong(hints *storage.SelectHints, matchers []*labels.
 	return (slices.Contains(rateFunctions, hints.Func) || hints.Func == "") && hints.Step != 0
 }
 
-// appendStaleMarker terminates an accelerated (substitute-backed) series with a
-// Prometheus stale marker one step past the last sample the SQL layer returned.
+// isSQLFilled reports whether the series was forward-filled 5m by
+// FillGapsPlanner. That is exactly the substitute-backed set (created only by
+// the vector_range/vector_agg optimizers, all of which route through the fill).
 //
-// The accelerated path builds a series with Prolong=false directly from
-// ClickHouse rows, and the SQL densifier (planner.fillGaps) already forward-fills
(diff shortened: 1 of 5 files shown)


mgai-rchung:
gigapipe maintainers: the failing test needs a change in the qryn-test repo. How do you coordinate these changes across the two repos?

reader/service/prom_queryable.go
```diff
@@ -210,6 +210,39 @@ func (c *CLokiQuerier) isProlong(hints *storage.SelectHints, matchers []*labels.
 	return (slices.Contains(rateFunctions, hints.Func) || hints.Func == "") && hints.Step != 0
```
rita7lopes, line 210: any other hints.Func -> false || Step == 0 -> false

reader/service/prom_queryable.go
rita7lopes, line 207: returns false for the only SQL-filled substitute-backed

rita7lopes (changes requested):
Hello @mgai-rchung,

Thank you very much for picking the issue and opening the PR, we really appreciate it!
In general the PR targets what I suggested on the issue, and the PR you opened on the gigapipe-tests repo correctly sorts the e2e tests.

On the PR description you added that `appendStaleMarker` applies to the "accelerated/substitute path (`Prolong=false`)". Those two aren't the same set, and the difference is where the problem is.
If you check my comment on the service/prom_queryable.go file, specifically:
`return (slices.Contains(rateFunctions, hints.Func) || hints.Func == "") && hints.Step != 0` the hints.Func is the enclosing function, and substitutes are only ever created by two optimizers: optimizer/vector_rango.go and optimizer/vector_agg.go
So these all reach `Select()` with `Prolong=false` and **no** substitute:

`abs`, `ceil`, `floor`, `sqrt`, `ln`, `exp`, `round`, `sgn`, `sort`, `timestamp`, `absent`, `scalar`, the trig functions, plus `topk`, `bottomk`, `quantile`, `histogram_quantile(q, m)`, `label_replace`, `label_join`, `clamp*`.

And those series are not SQL-filled. `FillGapsPlanner` is constructed in exactly one place — `planner/shared.go:73`, inside `bucketedValues` — reached only from `over_time.go:94` and `counter.go:25`. `HintsPlanner`, `ValuesPlanner` and the `Downsample*` planners never call it. `HintsPlanner` (`planner/hints.go:36`) does an `argMax(val, timestamp_ms)` regroup onto the step grid instead, which is a grouping, not a fill — the rows stop at the last real bucket.

Conclusion: for those functions the marker caps a 5m carry that was never applied, and the engine's own (correct, single) 5m lookback gets cut off instead. 

I think gatting the actual condition instead of on `!isProlong` is the simplest way to address this case. That also documents the invariant that makes the `+ step` offset correct: the rows were already carried 5m by SQL, so one step past the last row is the boundary. Right now that reasoning lives only in the comment.

A unit test covering one non-substitute instant-vector function (`abs` or `topk`) would be useful to make sure this scenario is covered. The current `appendStaleMarker` tests exercise the helper, but nothing asserts which series it gets applied to.

LMK what you think about this @mgai-rchung

mgai-rchung:
Thanks @rita7lopes, this is a great catch — you're right that `!isProlong` and "SQL-filled" are different sets, and the gap is exactly where the bug is.

I verified the flow you described:
- `FillGapsPlanner` is only constructed in `bucketedValues` (`planner/shared.go:73`), reached only from `over_time.go` and `counter.go`.
- Substitutes are only created by `optimizer/vector_range.go` and `optimizer/vector_agg.go`, and every planner they use (`OverTimePlanner`, `CounterPlanner`, `CounterFlagsPlanner`, and `AggPlanner` over `NewInstantVectorPlanner`) goes through `bucketedValues`.
- `HintsPlanner` (`planner/hints.go:36`) — used for `abs`/`ceil`/`topk`/`histogram_quantile`/`label_replace`/trig/etc. — does an `argMax(val, timestamp_ms)` regroup onto the step grid, not a fill, so those rows stop at the last real bucket.

So the old `!isProlong` gate wrongly appended a marker for those non-substitute instant-vector functions, capping a 5m carry that never happened and truncating the engine's own correct single 5m lookback.

Fixed in `71e4e40`:
- Added `isSQLFilled(matchers)` — true only when the `__name__` matcher resolves to a registered substitute — and gated `appendStaleMarker` on it at both call sites (replacing `!isProlong`).
- The doc comment now records the invariant that makes the `+step` offset correct: it's only valid because the SQL layer already carried the row 5m.
- Added unit tests asserting **which** series the marker applies to: true for a substitute-backed series, false for a non-substitute instant-vector function (`abs`/`topk`) and for a bare selector.

I'll also update the PR description to say `isSQLFilled` (substitute-backed) rather than "accelerated/substitute (`Prolong=false`)". Let me know if you'd like anything else.

reader/service/prom_queryable.go
mgai-rchung: fixed

reader/service/prom_queryable.go
mgai-rchung: fixed

rita7lopes:
@mgai-rchung thank you for the changes, merging this now!

rita7lopes (approved):

PR #953, opened 2026-08-31
all: apply go fix modernizers

janeblower:
Go 1.27 moved the gopls modernize analyzers into the toolchain, so `go fix`
now rewrites constructs that gained stdlib replacements over the last few
releases. One pass over the tree, so this churn does not land in feature
diffs later.

Commands: `go fix ./...`, then `gofmt -w .` (go fix does not reformat after
itself), then `go build ./...` and `go test ./...` — both exit 0, 28 packages
ok. Re-running `go fix -json ./...` afterwards confirms nothing actionable is
left.

171 diagnostics in handwritten code, all but five applied. The other 106 ar
in generated `.pb.go` files, untouched here.

| Analyzer | n | Change |
| --- | ---: | --- |
| `any` | 101 | `interface{}` to `any` |
| `embedlit` | 13 | Promoted fields set directly in composite literals |
| `mapsloop` | 9 | Copy loop to `maps.Copy` |
| `rangeint` | 8 | `for i := 0; i < n; i++` to `for i := range n` |
| `inline` | 5 | Call inlinable wrappers directly; 2 need type inference the tool lacks |
| `slicescontains` | 5 | Scan loop to `slices.Contains` |
| `stringsseq` | 4 | `range strings.Split` to `strings.SplitSeq` |
| `slicesbackward` | 4 | Reverse index loop to `range slices.Backward` |
| `minmax` | 3 | `math.Min`/`math.Max` and if-else to the builtins |
| `stringsbuilder` | 3 | `+=` accumulation to `strings.Builder` |
| `newexpr` | 3 | Left as is, see below |
| `atomictypes` | 2 | `atomic.AddInt32(&x, …)` to `atomic.Int32` |
| `forvar` | 2 | Drop the `x := x` loop copy, unneeded since 1.22 |
| `stringscutprefix` | 2 | `HasPrefix`+slice to `strings.CutPrefix` |
| `waitgroupgo` | 2 | `wg.Add(1)` + `go func()` to `wg.Go` |
| `errorsastype` | 1 | `errors.As` to `errors.AsType[T]` |
| `plusbuild` | 1 | Drop `// +build`, `//go:build` covers it |
| `reflecttypefor` | 1 | `reflect.TypeOf((*T)(nil)).Elem()` to `reflect.TypeFor[T]()` |
| `slicessort` | 1 | `sort.Slice` with a plain less func to `slices.Sort` |
| `stringscut` | 1 | `strings.Index` + two slices to `strings.Cut` |

Not everything here is straight `go fix` output. Three hand edits on top:

- **Layout in 5 files.** The rewrites leave a closing brace on the value line
  and blank lines where code was removed. gofmt does not fix either.
- **Reverted `newexpr` in 2 files.** It stamps the helper `//go:fix inline`,
  which is an instruction for a later pass, and it converted only one of
  `float64p`'s two calls, since `new(20)` is a `*int`.
- **`strings.Builder` in 2 files.** `stringsbuilder` emitted
  `b.WriteString("; " + r)`, concatenating right back. Split into two writes,
  and one became `fmt.Fprintf(&path, ".%d", *i)`.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,6 +16,9 @@ jobs:
       with:
         go-version-file: go.mod
 
+    - name: gofmt
+      run: test -z "$(gofmt -l .)" || { gofmt -l .; exit 1; }
+
     - name: Build Gigapipe
       run: |
         go mod tidy
--- a/ctrl/maintenance/shared.go
+++ b/ctrl/maintenance/shared.go
@@ -29,7 +29,7 @@ func ConnectV2(dbObject *config.ClokiBaseDataBase, database bool) (clickhouse_v2
 		MaxOpenConns:    10,
 		MaxIdleConns:    2,
 		ConnMaxLifetime: time.Hour,
-		Settings: map[string]interface{}{
+		Settings: map[string]any{
 			"allow_experimental_database_replicated": "1",
 			"materialize_ttl_after_modify":           "0",
 		},
--- a/ctrl/qryn/maintenance/update.go
+++ b/ctrl/qryn/maintenance/update.go
@@ -127,8 +127,8 @@ func getSQLFile(strContents string) ([]string, error) {
 	return res, nil
 }
 
-func getDBExec(db clickhouse.Conn, env map[string]string, logger logger.ILogger) func(query string, args ...[]interface{}) error {
-	return func(query string, args ...[]interface{}) error {
+func getDBExec(db clickhouse.Conn, env map[string]string, logger logger.ILogger) func(query string, args ...[]any) error {
+	return func(query string, args ...[]any) error {
 		name := fmt.Sprintf("tpl_%d", rand.Uint64())
 		tpl, err := template.New(name).Parse(query)
 		if err != nil {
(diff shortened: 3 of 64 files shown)


janeblower:
docs/contributing.md asks for gofmt, but nothing in CI checks it. Worth adding a
step to go.yml:

    - name: gofmt
      run: test -z "$(gofmt -l .)" || { gofmt -l .; exit 1; }

rita7lopes:
@janeblower from the merge of our other two PRs #954 and #955, now this one is smaller - please rebase onto main so we can review and merge this one too

rita7lopes:
> docs/contributing.md asks for gofmt, but nothing in CI checks it. Worth adding a step to go.yml:
> 
> ```
> - name: gofmt
>   run: test -z "$(gofmt -l .)" || { gofmt -l .; exit 1; }
> ```

I agree with this - can you add it as part of this PR?

rita7lopes (changes requested):
Resolve the conflicts and add the line to the CI action

janeblower:
@rita7lopes done!
Re-ran `go fix` on current master instead of rebasing and re-applied the manual fixes. 
gofmt step added to go.yml

rita7lopes (approved):
Thank you!!! Merging now @janeblower

PR #954, opened 2026-08-31
all: switch to math/rand/v2

janeblower:
The top-level math/rand/v2 functions are safe for concurrent use and need no seeding, so the per-struct *rand.Rand fields and the locks guarding them can go. Two of those locks guarded nothing else: the one in the writer service registry, and the global one in the pattern controller, which was taken on every single log line whenever pattern downsampling was enabled.

Two bugs fell out along the way:
 - writer/pattern/controller: random was declared but never assigned, so skipLine() dereferenced a nil *rand.Rand as soon as LogPatternsDownsampling was configured below 1.
 - CLokiQueriable.random was written and copied but never read — removed.

Mutexes that protect more than the generator (reader/registry/static.go, writer/service/generic_insert.go) are kept; only the lock around the rand call itself is dropped.

9 files, −29 lines. go build ./... and go test ./... pass.

### DIFF
--- a/ctrl/qryn/maintenance/update.go
+++ b/ctrl/qryn/maintenance/update.go
@@ -4,12 +4,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	rand2 "math/rand"
+	"math/rand/v2"
 	"regexp"
 	"strconv"
 	"strings"
 	"text/template"
-	"time"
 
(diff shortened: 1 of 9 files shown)


rita7lopes (approved):
Good one @janeblower !
Merging

PR #955, opened 2026-08-31
Dead code

janeblower:
# Remove dead packages and stale replace directives

Four packages with no importer anywhere in the tree, plus five `replace`
directives that resolve nothing. 22 files, -1326 lines, no behaviour change.

Drops **6 direct** dependencies — `gofiber/fiber/v3`, `labstack/gommon`,
`go-kit/kit`, `c2h5oh/datasize`, `gopkg.in/yaml.v2`, `valyala/bytebufferpool` —
and **10 indirect**, including `valyala/fasthttp`, the second HTTP stack in the
graph. `go build ./...` and `go test ./...` pass on each of the four commits
separately.

| Removed | Why | Born | Died |
|---|---|---|---|
| `writer/utils/helpers/limited_buffer.go` | `GetRaw*` take a `*fiber.Ctx`; the pools `SetGlobalLimit` sizes are read only from those | `7580d08`, 2025-04-07 | never called |
| `writer/utils/httpresponse/` | `CreateBadResponse(*fiber.Ctx, …)`, incompatible with the `net/http` layer | `7580d08`, 2025-04-07 | `f907eda`, 2025-07-15 — controllers moved off fiber |
| `reader/utils/logger/echologrus/` | logrus adapter for echo's logger interface; echo has never been a dependency | `7580d08`, 2025-04-07 | never imported |
| `writer/utils/util/` (16 files) | Loki/Cortex helpers — ring tokens, tenant runtime config, bucketed byte pool; gigapipe has no ring, no ingesters, no per-tenant runtime config | `7580d08`, 2025-04-07 | never imported |
| 4 `replace` directives | `go mod why`: "main module does not need module" | `7580d08`, 2025-04-07 | 3 arrived dead; `cloud.google.com/go/compute` left the graph in `f899eef`, 2025-10-30 |
| `replace k8s.io/client-go v12.0.0+incompatible` | pinned version absent from the graph; the module resolves to v0.36.3 via `prometheus/util/logging` | `7580d08`, 2025-04-07 | `35ea45b`, 2025-11-03 |

`go.sum` does not move on the `replace` commit, which is the proof that none of
the five was doing anything.

What `writer/utils/util` had worth keeping is in the standard library now:
`StringSliceContains` > `slices.Contains`, `TokenFor` > `hash/maphash`,
`NewSizeReader` > a counter over `io.Reader`. `writer/utils/helpers/size_getter.go`
(21 callers) stays.

## The input buffer limit

`helpers.SetGlobalLimit(config.HTTP_SETTINGS.InputBufferMB * 1024 * 1024)` in
`writer/plugin/qryn_writer.go` was the only live line among the deleted code. It
sized `requestPool` and `pbPool`, which are read only from the dead `GetRaw*`
path, so `HTTP_SETTINGS.InputBufferMB` (default 200) has capped nothing since
the fiber controllers went away in July 2025. Removing the call does not change
what any request is allowed to do — it only makes the situation visible.

The option itself lives in `metrico/cloki-config` and is untouched here.
Handlers that need a cap already set their own: `writer/controller/otlp_metrics.go:75`
and `writer/utils/unmarshal/go_pprof.go:267` use `io.MaxBytesReader`. Putting a
global limit back on the `net/http` path would be a behaviour change for anyone
currently posting more than 200 MB, so it belongs in its own PR if wanted.

## Note

The module is public (`github.com/metrico/qryn/v5`), so removing exported
packages is formally breaking under SemVer. None of them is documented or
referenced outside its own sources — a minor release with a changelog line
seems right.


rita7lopes (approved):
Thanks for this @janeblower 
Merging!

PR #956, opened 2026-09-01
fix(otlp-profiles): name unsymbolized frames and fall back for service name

rita7lopes:
## Summary

Gigapipe accepts OTLP profiles from the [OpenTelemetry eBPF profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler) — verified end to end by running the real agent against a local instance. But two consumer-side defects made the result practically unusable. Both threw away data that had already arrived on the wire.

## 1. Unsymbolized native frames had no identity

The eBPF profiler symbolizes interpreted runtimes in-agent and leaves native frames to the backend, sending an address plus a Mapping. `frameName` dropped the mapping and emitted `+0x<address>`.

Measured against the real agent: **60.5% of frames arrive unsymbolized** (3818 of 6312 locations), so most of a flamegraph rendered as anonymous hex. 99.8% of those carried a mapping filename *and* a build id — all discarded.

Frames are now named `<filename>@<build id>+0x<offset>`:

- the address is a **file-relative offset**, not a runtime virtual address, so the name is stable across processes and runs and merges correctly
- the filename is used exactly as the producer sends it, no normalisation
- the 8-char build id separates same-named binaries from different images — the case the filename alone cannot disambiguate
- each component degrades independently: no build id → `<filename>+0x<offset>`, no resolvable mapping → `+0x<offset>`

Result on real agent output: unsymbolized frames carrying identity went **0% → 99.5%**.

## 2. Every profiled process collapsed into one service

The profiler sets `service.name` only for APM-linked processes. `extractOTLPMeta` read that attribute or fell back to the literal `"unknown_service"`.

Measured: **90 series, 100% `unknown_service`** — while the tags cleanly separated `runc`, `wget`, `dockerd`, `clickhouse`, `python3.12`, `containerd` and others.

Service name now resolves through the candidate chain the **traces path already used**, which already ended in `process.executable.name`. Result: **1 service → 12**. The 8 series still resolving to `unknown_service` are kernel threads, which genuinely have no executable name.

## 3. One resolver, not one per signal

Rather than copy the candidate list into the profiles path, the list and the resolution both moved into `otlp_service_name.go`. Traces and profiles share the same list and the same function, differing only in how they look an attribute up — pdata `pcommon.Map` for profiles, protobuf `KeyValue` slice for traces. A process cannot now land under two different service names depending on which signal it emitted, which is the point of a shared name.

This shortened the traces path from 30 lines to 12 with no behaviour change — the pre-existing `TestOtlpGetServiceNamesUsesFirstNonEmptyPreferredAttribute` passes unmodified.

## Performance

`frameName` now resolves **once per distinct location** rather than once per frame occurrence. That more than pays for the added build-id lookup:

| | 100k-frame profile |
|---|---|
| tree construction, before | 40.98 ms |
| tree construction, after | ~35 ms |

Storage grows **+20% in raw frame-name bytes**, which compresses to **+11.6% under LZ4** (ClickHouse's default) and +2.4% under ZSTD — roughly **+1% of on-disk table size**. Nothing parses the name format.

The read path is not untouched, and calling it that was wrong. `PlanMergeTraces` -> `MergeRawPlanner` reads the `tree`/`functions` columns and so shows the new names, but `PlanMergeProfiles` -> `otlpToPProf` rebuilds locations from the stored payload and never consults the mapping table — so the same profile renders as `libc.so.6@a3f2c1d4+0x1234` in a flamegraph and as an anonymous location in a pprof export. `sliceOTLPProfile` already retains the filenames and build ids a fix needs, so this is a gap to close in a dedicated PR, not a blocker here. Thanks to @44YHC for catching it.

## Behaviour changes worth flagging

- An attribute present but **empty** no longer erases the resolved name — it is treated as absent, matching the traces path. Previously `service.name=""` produced an empty service name.
- **Resource attributes now outrank scope attributes**, matching OTel resource semantics. Previously scope silently overwrote resource.
- `Location.mapping_index` is a plain `int32` with no presence bit, so index 0 is indistinguishable from unset. We take the spec-correct reading — index 0 *is* mapping 0, since the schema offers no way to say "no mapping". Documented in `frameName`, including why a heuristic guard would be the worse trade.

## Operator note

Profiles that previously resolved to `unknown_service` get new service names and therefore **new series fingerprints**. Already-stored frames keep their old `+0x...` names and will not merge with newly-ingested ones. No migration required, but the discontinuity is visible on dashboards spanning the deploy.

## Testing

- `go build ./...`, `go vet ./writer/...`, 25 test packages green
- new table-driven tests: build-id priority and degradation, service-name fallback and precedence, tag retention
- full `make e2e-ci`: **175/175, snapshots 126/126**
- live verification against the real eBPF profiler on kernel 6.12.54

⚠️ **Requires metrico/gigapipe-tests#17 to merge first** — e2e snapshots currently assert the old `+0x2a` naming. That PR also adds a regression guard for the cross-binary case this fix exists for.


writer/utils/unmarshal/otlp_profile.go
```diff
@@ -317,6 +423,23 @@ func buildOTLPTree(p pprofile.Profile, dict pprofile.ProfilesDictionary) (
 	locs := dict.LocationTable()
 	stacks := dict.StackTable()
 
+	// A location repeats across every sample whose stack walks through it, and
+	// resolving its name costs several dictionary lookups (function table, or
+	// mapping table plus a scan of the attribute table for the build id). Resolve
+	// each distinct location once instead of once per frame occurrence.
+	names := make(map[int32]string, locs.Len())
```
44YHC: **Blocking:** this pre-sizing is a real regression on batched exports.

`locs` is the **export-wide** dictionary (`dict := profs.Dictionary()` in `decodeProfiles`), but `buildOTLPTree` runs **once per profile**. So every profile in a batch allocates a map pre-sized for the whole batch's location table — and the eBPF profiler is exactly the producer that sends many small profiles under one shared dictionary (this PR's own numbers: 90 series, 6312 locations).

Measured on a batch of 90 profiles x 20 frames over a 6312-location dictionary:

| | ns/op | B/op |
|---|---|---|
| master | 677,515 | 619,221 |
| this PR | 2,769,744 | 20,337,243 |
| this PR, size hint dropped | 982,560 | 787,870 |

4x slower and 33x more memory than master for this shape, entirely from the hint.

To be fair to the change: my synthetic shape gives each location one visit per profile, which is the worst case for the cache and the best case for master — the intra-profile win you measured on the 100k-frame profile is genuine. The pre-sizing waste is independent of that and is a straight loss.

Dropping `, locs.Len()` recovers most of it. But the better fix, and the one that actually matches this comment's stated goal of resolving each distinct location once, is to hoist the cache up to `decodeProfiles` and pass it in. The dictionary is shared and immutable across the batch, so a batch-wide cache is valid — and on real data (many processes sharing libc/kernel mappings) that turns N-profiles-x-per-location into genuinely once per export.

writer/utils/unmarshal/otlp_profile.go
```diff
@@ -302,8 +352,64 @@ func frameName(loc pprofile.Location, dict pprofile.ProfilesDictionary) string {
 			}
 		}
 	}
-	buildID := ""
-	return fmt.Sprintf("%s+0x%x", buildID, loc.Address())
+	binary, build := "", ""
```
44YHC, line 404: Not a blocker and not introduced here, but worth a tracked follow-up: the description says "read path is untouched", which reads as reassurance when it's really a gap.

Both read paths are user-visible:

- `PlanMergeTraces` -> `MergeRawPlanner` reads the `tree`/`functions` columns -> **gets the new names**.
- `PlanMergeProfiles` -> `otlpToPProf` (`reader/service/otlp_pprof.go`) rebuilds locations from the stored payload and never touches the mapping table -> **unsymbolized frames still have no function at all**.

So the same profile renders as `libc.so.6@a3f2c1d4+0x1234` in the flamegraph and as an anonymous location in a pprof export.

Good news: `sliceOTLPProfile` already seeds `refMap` and calls `addAttrIdxs(m.AttributeIndices())` in steps 3-4, so the stored payload retains the filenames and build ids a follow-up would need. Nothing to change here.

writer/utils/unmarshal/otlp_profile.go
```diff
@@ -302,8 +352,64 @@ func frameName(loc pprofile.Location, dict pprofile.ProfilesDictionary) string {
...
+const buildIDLen = 8
+
+// shortBuildID returns the mapping's preferred build id, truncated, or "" when
+// the mapping carries none.
+//
+// Preference runs over buildIDKeys, NOT over the order the producer happened to
+// append the attributes in: a mapping carrying several build ids must always
+// resolve to the same one, or the same frame would be named differently
+// depending on attribute ordering and would stop merging with itself.
+func shortBuildID(mp pprofile.Mapping, dict pprofile.ProfilesDictionary) string {
```
44YHC: Nit: `shortBuildID` is called once per *location*, but a build id is a property of the *mapping*, and many locations share one mapping. Caching by mapping index would drop the repeated attribute-table scans entirely.

Minor next to the `names` map issue, but it falls out naturally if you're touching the caching anyway.

writer/utils/unmarshal/otlp_service_name.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// There is deliberately ONE list for every OTLP signal. Traces resolve a span's
+// service from it (otlpGetServiceNames) and profiles resolve a profile's
+// service from it (extractOTLPMeta). Keeping a copy per signal would let them
+// drift, and the same process would then land under two different service names
+// depending on which signal it emitted — which breaks exactly the correlation
+// (this trace, and the profile of the process that served it) the shared name
+// exists to provide.
+//
+// Add a key here, not in a caller.
```
44YHC: This invariant isn't actually enforced repo-wide: `reader/service/parse_otlp_json.go:33` still has a byte-for-byte copy of the old `otlpGetServiceNames`, with both candidate lists inlined.

It's a different package (`reader/service` vs `writer/utils/unmarshal`), so unifying means putting the list somewhere both can reach — `shared/otlp` is the natural home, and it's already where `ProfilePayloadType` lives for exactly this reason.

As it stands the comment asserts an invariant the repo doesn't hold, which is worse than not claiming it. Either move the list to `shared/otlp` and update the reader, or soften the claim to describe the writer package only.

writer/utils/unmarshal/otlp_service_name.go
```diff
@@ -0,0 +1,48 @@
...
+	"faas.name",
+	"k8s.deployment.name",
+	"process.executable.name",
+}
+
+// otlpLocalServiceNameKeys is otlpServiceNameKeys with peer.service ahead of
+// it. It applies only when resolving the LOCAL service of a span, where an
+// explicit peer.service names the service the span is talking to and therefore
+// outranks the resource's own identity.
+var otlpLocalServiceNameKeys = append([]string{"peer.service"}, otlpServiceNameKeys...)
```
44YHC, line 43: Two things here.

**The rationale is backwards.** `peer.service` is by definition the *remote* service, so "it names the service the span is talking to, therefore it outranks the resource's own identity **as the local name**" doesn't follow. The ordering is pre-existing — master does the same — and preserving it in a refactor is the right call. But the new comment presents a long-standing oddity as deliberate design. Per the repo's comment convention (state the rule the code follows, not a rationale), I'd either drop the justification and just say `peer.service` leads the local list, or fix the ordering in a separate change where the behaviour shift is visible.

**Minor:** `append([]string{...}, otlpServiceNameKeys...)` copies at init, so a later append to `otlpServiceNameKeys` silently won't reach the local list — quietly breaking the "Add a key here, not in a caller" instruction three lines above. Worth a word, or build both lists from a shared base.

writer/utils/unmarshal/otlp_profile_test.go
```diff
@@ -641,3 +641,242 @@ func TestSliceOTLPProfilePrunesAndRoundTrips(t *testing.T) {
...
+	const want = "/usr/bin/busybox+0x3107"
+	for _, f := range functions {
+		if f.ValueStr == want {
+			return
+		}
+	}
+	t.Fatalf("expected %q, got %+v", want, functions)
+}
+
+// buildIDProfile builds a one-native-frame profile whose mapping carries the
```
44YHC: Nit: stale name — the function is `buildIDFrameName`, not `buildIDProfile`.

writer/utils/unmarshal/otlp_profile_test.go
```diff
@@ -641,3 +641,242 @@ func TestSliceOTLPProfilePrunesAndRoundTrips(t *testing.T) {
...
+			if got := buildIDFrameName(t, tt.filename, tt.attrs); got != tt.want {
+				t.Fatalf("got %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+// metaForAttrs runs extractOTLPMeta over a profile carrying the given resource
+// and scope attributes, and returns the resolved metadata.
+func metaForAttrs(resAttrs, scopeAttrs map[string]string) otlpProfileMeta {
```
44YHC, line 800: `scopeAttrs` is passed `nil` at every call site, so the parameter is currently dead — and with it, the resource-vs-scope precedence change goes untested.

That's one of the three behaviour changes the description explicitly flags ("Resource attributes now outrank scope attributes, matching OTel resource semantics"), and it's the one most likely to surprise someone. Two cases would cover both branches of the new `lookup` closure:

- same key on resource and scope with different values -> resource wins
- key empty on resource, non-empty on scope -> falls through to scope

Also worth one for a non-string attribute: traces read via `GetStringValue()` (non-string -> `""` -> treated as absent) while profiles use `AsString()` (stringifies ints), so a non-string `process.executable.name` resolves differently across the two signals. Theoretical, but it's the exact drift the shared resolver exists to prevent.

44YHC (commented):
Reviewed the diff and ran the package on the PR head in a scratch worktree: `go vet ./writer/utils/unmarshal/...` clean, `go test ./writer/utils/unmarshal/...` green.

**Verdict: sound fix, well-argued, with one measurable performance regression I'd want fixed before merge.**

The diagnosis is correct and the numbers in the description are credible. I verified the `mapping_index` claim against pdata `pprofile@v0.159.0` — `Location.MappingIndex()` really is a bare `int32` with no `Has*` accessor, so the spec-correct reading the PR takes is the only one available. The traces refactor is behaviour-preserving as claimed.

Blocking finding is inline on `buildOTLPTree` (the `names` map pre-sizing — 4x slower and 33x more memory than master on a batched export, measured). Everything else is a suggestion or a nit.

### Test coverage

New tests are well shaped — table-driven, good names, comments that explain *why* each case matters rather than restating the assertion. Three gaps:

- **"Resource attributes now outrank scope attributes" is untested**, and it's one of the three behaviour changes the description explicitly flags. See the inline note on `metaForAttrs`.
- No coverage for `frameName` degrading to a bare `+0x<offset>` when the mapping index is out of range.
- No coverage for a location that *has* lines but whose function name resolves empty, falling through to the mapping fallback.

### Risks

The operator note is right and important — new fingerprints for profiles that previously resolved to `unknown_service`, and stored `+0x...` frames that won't merge with newly-ingested named ones. Real but bounded, and correctly disclosed up front. The `metrico/gigapipe-tests#17` ordering dependency is also correctly flagged.

No security concerns: the frame name is producer-controlled string data going into an existing ClickHouse column, and nothing parses the format.

akvlad (approved):

writer/utils/unmarshal/otlp_profile.go
rita7lopes: Confirmed and fixed in faab0a7. You are right about the mechanism: `dict` is export-wide, `buildOTLPTree` runs per profile, so the `locs.Len()` hint allocated a whole-dictionary map for every profile in a batch.

Took the second option — the cache is now an `otlpFrameNamer` built once per export in `decodeProfiles` and passed down. That also makes the code match what the comment already claimed. Neither map is pre-sized now: a profile commonly touches a small fraction of a large shared dictionary, so the length of the table is the wrong hint even at export scope.

Re-measured rather than assume, 20 iterations over a 6312-location dictionary:

| shape | master | previous commit | now |
|---|---|---|---|
| 90 profiles, no location reuse | 1.56 ms / 622 KB | 7.38 ms / 20.4 MB | 4.06 ms / 900 KB |
| 90 profiles, shared locations | 28.6 ms / 12.0 MB | 44.3 ms / 33.9 MB | 26.3 ms / 12.4 MB |
| 1 profile, 100k frames | 44.9 ms | 32.9 ms | 33.7 ms |

Your shape is row 1, and your caveat about it is the right one: every location is visited exactly once there, so the cache can never hit and only the build-id lookup is paid. Memory goes from 33x master to 1.4x; the residual time gap is the feature itself, not the caching, and it closes as soon as locations repeat. Row 2 is the whole-machine shape this PR targets — now 8% faster than master at +3% memory.

writer/utils/unmarshal/otlp_profile.go
rita7lopes: Fixed in faab0a7, and you are right that it falls out of the cache work. `shortBuildID` is now memoized by mapping index and the pure scan moved to `scanBuildID`, so a build id is resolved once per mapping per export instead of once per location.

writer/utils/unmarshal/otlp_profile_test.go
rita7lopes: Fixed in faab0a7.

writer/utils/unmarshal/otlp_profile_test.go
rita7lopes: Fixed in faab0a7 — `scopeAttrs` is no longer dead.

Added `TestExtractOTLPMetaResourceOutranksScope` with four cases: same key on both (resource wins), key empty on resource (falls through to scope), key absent from resource (falls through to scope), and a higher-priority key on the scope beating a lower-priority key on the resource — that last one pins the interaction between the two orderings, which is the part most likely to be broken by a later edit.

Also took the non-string case: `TestExtractOTLPMetaStringifiesNonStringCandidate` pins that profiles stringify via `AsString()` where traces treat a non-string as absent via `GetStringValue()`. Left as a documented difference rather than aligned — both signals share the key list and the resolution order, and only the lookup differs because each reads the representation its own decoder produced. Aligning them would change trace behaviour, which belongs with the `shared/otlp` unification.

Your other two gaps are covered too: `TestFrameNameOutOfRangeMappingKeepsOffset` and `TestFrameNameEmptyFunctionNameFallsBackToMapping`.

writer/utils/unmarshal/otlp_service_name.go
rita7lopes: You are right, and the rationale was mine to get wrong — `peer.service` names the remote service by convention, so "therefore it outranks the resource's own identity as the local name" does not follow.

Took your first option in faab0a7: the comment now states that `peer.service` leads the local list, that the ordering is pre-existing, and that changing it would re-attribute already-stored spans, so it belongs in a commit where that behaviour change is the visible point. No justification invented for it.

On the `append`: noted rather than restructured, because it is narrower than it looks. The copy is made at init from the final value of the literal, so a key added to `otlpServiceNameKeys` does reach the local list — only a *runtime* append would be lost, and nothing does one. The comment now says exactly that, and says to build both from a common base if that ever changes.

writer/utils/unmarshal/otlp_service_name.go
rita7lopes: Confirmed — `reader/service/parse_otlp_json.go:33` really is a byte-for-byte copy with both lists inlined, so the comment asserted something the repo does not hold.

Softened it in faab0a7 rather than unified now. The comment names the reader copy explicitly and states the two things unification has to reconcile beyond moving the list: the reader accepts only string-typed attribute values (`GetStringValue()`), and falls back to `OTLPResourceNoServiceName` rather than `unknown_service`. Both of those are read-path behaviour changes that deserve their own visible diff rather than riding along in a profiles-ingest fix.

The `shared/otlp` move is the right destination and it will come as a dedicated PR.

writer/utils/unmarshal/otlp_profile.go
rita7lopes: Agreed on both counts, and thanks for tracing it — I had not walked `PlanMergeProfiles` and the description sentence was reassurance I had not earned. Fixing the wording.

Confirmed the asymmetry is real: the flamegraph reads the `tree` column and gets the new names, while `otlpToPProf` rebuilds locations from the stored payload and never touches the mapping table, so one profile renders two different ways. Also confirmed your good news — `sliceOTLPProfile` already seeds `refMap` and calls `addAttrIdxs(m.AttributeIndices())`, so the filenames and build ids a fix needs are retained in the payload.

Nothing changes here; dedicated PR.

44YHC (approved):

PR #959, opened 2026-09-02
Replace hand-rolled helpers with the stdlib equivalents

janeblower:
## What

**`all: replace hand-rolled generic helpers with stdlib`** — three helpers that
predate the stdlib functions covering them are gone: `Unwrap` → `errors.AsType`,
`removeInPlace` → `slices.DeleteFunc`, `fastFillArray` → `slices.Repeat`. Plus
`x/exp/slices` → `slices`, one `pkg/errors.Errorf` → `fmt.Errorf`, and
`math.Min(math.Max(...))` → `min(max(...))`.

**`all: sort with slices.SortFunc`** — all 21 `sort.Slice`/`sort.Strings` call
sites move to `slices.SortFunc`/`slices.Sort`. The `sort` package is no longer
imported anywhere in the tree.

`go.mod` loses two direct dependencies: nothing else in the build graph imports
`golang.org/x/exp` or `github.com/pkg/errors`.

## Why

`slices.SortFunc` is the same pdqsort as `sort.Slice` — `sort/zsortfunc.go` and
`slices/zsortanyfunc.go` are line-for-line identical once you strip the naming —
but it skips the `reflect.Swapper` indirection and the reflection setup it
allocates for. Same for the rest: the stdlib versions are the same algorithms
without the wrapper, and `errors.AsType` avoids the `errors.As` reflection path
entirely.

The one place where the rewrite is not mechanical is the series comparator in
`CLokiQuerier.Select`. It called `LabelsArray()` on the right-hand series once
per label rather than once per comparison, and every call rebuilds and re-sorts
the label slice. Series of one metric share their leading labels, so that inner
call ran for almost every label of every comparison.

## Numbers

Microbenchmarks reproducing each changed shape, go1.27.0 linux/amd64:

| shape | before | after |
|---|---|---|
| series sort, 200 series × 8 labels sharing a prefix | 1269 µs, 14006 allocs | **353 µs, 3044 allocs** |
| sort 10k profile ids, descending | 703 µs, 2 allocs | **573 µs, 0 allocs** |
| sort 20 metric label pairs | 209 ns, 3 allocs | **90 ns, 0 allocs** |
| unwrap a wrapped `IQrynError` | 99 ns, 1 alloc | **22 ns, 0 allocs** |
| filter 10k profile nodes, sanitize-shaped predicate | 132 µs | 139 µs |
| fill 1000 sample-type bytes | 75 ns | 113 ns |

The last two get slower. `slices.DeleteFunc` clears the tail it drops so the
removed pointers stop being reachable, and the 5% is inside the predicate's own
map writes anyway. `slices.Repeat` runs the same doubling copy and the 38 ns is
noise next to the parsing around it; it also returns an empty slice for a zero
length where `fastFillArray` indexed `res[0]` and panicked, which `onEntries`
could reach with no timestamps.

## Behaviour

Sort order is unchanged everywhere the old boolean comparator was a strict
order — same algorithm, same permutation, verified over ~2400 randomised runs
across sizes and tie densities. Two comparators were not strict orders:

- `Select` returned `true` for fully equal label sets, claiming both `a < b` and
  `b < a`. `slices.CompareFunc` returns 0. Unreachable in practice:
  `ReshuffleSeries` merges equal label sets before the sort.
- The tempo metrics bucket sort compared `float64` with `<`, which is not an
  ordering when a value is NaN — `sort.Slice` could leave the slice unsorted.
  `cmp.Compare` gives a total order.

`clone` in `prof_merge_v1.go` is deliberately kept. `new(*v)` is the same shallow
copy, but without the type parameter `copylocks` starts reporting the
`protoimpl.MessageState` copy the helper was hiding. That hazard predates this
change and deserves its own fix (`proto.Clone` is not a drop-in: the callers
re-clone only some fields on purpose).

`go build`, `go vet` (no new findings) and `go test ./...` pass — 26 packages.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/kr/logfmt v0.0.0-20210122060352-19f9bcb100e6
 	github.com/lestrrat-go/file-rotatelogs v2.4.0+incompatible
 	github.com/metrico/cloki-config v0.0.96
-	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.24.1
 	github.com/prometheus/common v0.70.1
 	github.com/prometheus/prometheus v0.314.0
@@ -35,7 +34,6 @@ require (
 	go.opentelemetry.io/collector/pdata v1.65.0
 	go.opentelemetry.io/collector/pdata/pprofile v0.159.0
 	go.opentelemetry.io/proto/otlp v1.11.0
-	golang.org/x/exp v0.0.0-20260709172345-9ea1abe57597
 	golang.org/x/sync v0.22.0
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260724162435-b2f20204f0df
 	google.golang.org/grpc v1.85.0-dev
@@ -91,6 +89,7 @@ require (
 	github.com/paulmach/orb v0.13.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.8 // indirect
 	github.com/pierrec/lz4/v4 v4.1.27 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/prometheus/procfs v0.21.1 // indirect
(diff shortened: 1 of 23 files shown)


janeblower:
Unrelated failure: qryn-test landed a new snapshot ahead of the gigapipe codefor it, so e2e is red on master too.

rita7lopes (approved):
Very nice one @janeblower !
As a follow up, if you feel like it, Select's comparator still calls LabelsArray() twice per comparison, and labelsGetter.Get (prom_queryable.go:462-478) allocates and re-sorts on every call. Precomputing the key once per series would be interesting to see what improvement it can give us, as a new row on your table comparison 😊

PR #960, opened 2026-09-03
smart_buffer: grow the RAM chunk on demand

janeblower:
The reader wraps every response in a **SmartBuffer**, which preallocated its full 5MB RAM chunk in the constructor. All ten reader endpoints go through it, including **labels**, **label/{name}/values** and **series** — the ones Grafana polls constantly with replies of a few hundred bytes. Each of those requests malloc'd and page-faulted 5MB to hold 200 bytes.

Fix: start the chunk at 8KB and let **append** grow it. The 5MB spill-to-disk threshold is unchanged — it's checked against the byte count written, not the slice capacity. Small responses: 511µs → 1.4µs, 5MB → 8KB per request.

### DIFF
--- a/reader/utils/smart_buffer/ram_chunk.go
+++ b/reader/utils/smart_buffer/ram_chunk.go
@@ -8,8 +8,13 @@ import (
 // ErrBufferFull is returned when the RAM chunk cannot accept more data.
 var ErrBufferFull = errors.New("buffer is full")
 
-// chunkSize defines the maximum size of the RAM buffer (5MB).
-const chunkSize = 5 * 1000 * 1000 // 5MB
+const (
+	// chunkSize defines the maximum size of the RAM buffer (5MB).
+	chunkSize = 5 * 1000 * 1000 // 5MB
+	// initialChunkSize is the starting capacity; append grows it as needed,
+	// so a small response never touches the whole 5MB.
+	initialChunkSize = 8 * 1024
+)
 
 // ramChunk is a fixed-size RAM buffer that accumulates data up to chunkSize.
 // It returns ErrBufferFull when it cannot accept more data.
@@ -18,10 +23,10 @@ type ramChunk struct {
 	size int
 }
 
-// newRAMChunk creates a new RAM chunk with pre-allocated capacity.
+// newRAMChunk creates a new RAM chunk that grows on demand up to chunkSize.
 func newRAMChunk() *ramChunk {
 	return &ramChunk{
-		data: make([]byte, 0, chunkSize),
+		data: make([]byte, 0, initialChunkSize),
 	}
 }
 
--- a/reader/utils/smart_buffer/ram_chunk.go
+++ b/reader/utils/smart_buffer/ram_chunk.go
@@ -11,43 +11,86 @@ var ErrBufferFull = errors.New("buffer is full")
 const (
 	// chunkSize defines the maximum size of the RAM buffer (5MB).
 	chunkSize = 5 * 1000 * 1000 // 5MB
-	// initialChunkSize is the starting capacity; append grows it as needed,
-	// so a small response never touches the whole 5MB.
-	initialChunkSize = 8 * 1024
+	// initialBlockSize is the size of the first block, small enough that a
+	// few-hundred-byte response costs a single small allocation.
+	initialBlockSize = 8 * 1024
+	// maxBlockSize caps the block size at Go's largest size class, so every
+	// block stays on the small-object allocation path.
+	maxBlockSize = 32 * 1024
 )
 
-// ramChunk is a fixed-size RAM buffer that accumulates data up to chunkSize.
+// ramChunk accumulates up to chunkSize bytes in a list of separately allocated
+// blocks. Blocks are never reallocated or copied, so the chunk allocates only
+// what is actually written: a growing single slice would copy everything
+// accumulated so far on each growth step. Blocks are reused after Clear.
 // It returns ErrBufferFull when it cannot accept more data.
 type ramChunk struct {
-	data []byte
-	size int
+	blocks    [][]byte
+	cur       int // block currently being filled
+	size      int
+	readBlock int
+	readOff   int
 }
 
-// newRAMChunk creates a new RAM chunk that grows on demand up to chunkSize.
+// newRAMChunk creates a new RAM chunk. No memory is reserved until the first write.
 func newRAMChunk() *ramChunk {
-	return &ramChunk{
-		data: make([]byte, 0, initialChunkSize),
-	}
+	return &ramChunk{}
 }
 
 // Write writes data to the RAM chunk, writing as much as will fit.
 // Returns the number of bytes written and ErrBufferFull if not all data could be written.
 func (r *ramChunk) Write(p []byte) (int, error) {
-	available := chunkSize - r.size
-	if available == 0 {
-		return 0, ErrBufferFull
+	written := 0
+	for len(p) > 0 {
+		if r.size >= chunkSize {
+			return written, ErrBufferFull
+		}
+		if r.cur == len(r.blocks) {
+			r.blocks = append(r.blocks, make([]byte, 0, r.nextBlockSize()))
+		}
+		block := r.blocks[r.cur]
+		if len(block) == cap(block) {
+			r.cur++
+			continue
+		}
+		n := min(len(p), cap(block)-len(block))
+		r.blocks[r.cur] = append(block, p[:n]...)
+		p = p[n:]
+		r.size += n
+		written += n
 	}
-
-	written := min(len(p), available)
-
-	r.data = append(r.data, p[:written]...)
-	r.size += written
-
-	if written < len(p) {
-		return written, ErrBufferFull
-	}
-
 	return written, nil
+}
+
+// nextBlockSize doubles the previous block size up to maxBlockSize, clamped to
+// the room left in the chunk so the blocks tile chunkSize exactly.
+func (r *ramChunk) nextBlockSize() int {
+	size := initialBlockSize
+	if r.cur > 0 {
+		size = min(cap(r.blocks[r.cur-1])*2, maxBlockSize)
+	}
+	return min(size, chunkSize-r.size)
+}
+
+// Read implements io.Reader over the accumulated blocks.
+// Returns io.EOF once all data has been read.
+func (r *ramChunk) Read(p []byte) (int, error) {
+	n := 0
+	for n < len(p) && r.readBlock < len(r.blocks) {
+		block := r.blocks[r.readBlock]
+		if r.readOff == len(block) {
+			r.readBlock++
+			r.readOff = 0
+			continue
+		}
+		copied := copy(p[n:], block[r.readOff:])
+		r.readOff += copied
+		n += copied
+	}
+	if n == 0 {
+		return 0, io.EOF
+	}
+	return n, nil
 }
 
 // Flush writes all accumulated data in the chunk to the provided writer
@@ -56,23 +99,24 @@ func (r *ramChunk) Flush(w io.Writer) error {
 	if r.size == 0 {
 		return nil
 	}
-	if _, err := w.Write(r.data); err != nil {
-		return err
+	for _, block := range r.blocks {
+		if _, err := w.Write(block); err != nil {
+			return err
+		}
 	}
 	r.Clear()
 	return nil
 }
 
-// Clear resets the chunk to empty state.
+// Clear resets the chunk to empty state, keeping the blocks for reuse.
 func (r *ramChunk) Clear() {
-	r.data = r.data[:0]
+	for i := range r.blocks {
+		r.blocks[i] = r.blocks[i][:0]
+	}
+	r.cur = 0
 	r.size = 0
-}
-
-// Bytes returns the current data in the chunk.
-// The returned slice should not be modified.
-func (r *ramChunk) Bytes() []byte {
-	return r.data
+	r.readBlock = 0
+	r.readOff = 0
 }
 
 // Size returns the current number of bytes stored in the chunk.
--- a/reader/utils/smart_buffer/ram_chunk_test.go
+++ b/reader/utils/smart_buffer/ram_chunk_test.go
@@ -1,0 +1,112 @@
+package smart_buffer
+
+import (
+	"bytes"
+	"io"
+	"testing"
+)
+
+func writeChunks(t *testing.T, r *ramChunk, total, per int) {
+	t.Helper()
+	data := make([]byte, per)
+	for written := 0; written < total; {
+		n, err := r.Write(data[:min(per, total-written)])
+		if err != nil {
+			t.Fatalf("Write() error = %v", err)
+		}
+		written += n
+	}
+}
+
+func TestRAMChunk_BlocksTileChunkSize(t *testing.T) {
+	t.Parallel()
+
+	r := newRAMChunk()
+	writeChunks(t, r, chunkSize, 4096)
+
+	if r.Size() != chunkSize {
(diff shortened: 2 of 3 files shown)


rita7lopes (changes requested):
Thanks @janeblower !!
The PR makes sense for the small queries indeed, however, the cases we need to be concerned (the most) are the heavy queries like `query_range`.
Doing a test with this code for a 8MB chunck the results show some problems when comparing with main's code:
Image

The problem is that the append regrows ~30 times and makes it all worse.

So, to get the best of both worlds, let's keep the 8KB initial chunk size, BUT, as soon as we have the first append overflow we jump straight to chunkSize instead of letting append ladder up.

janeblower:
@rita7lopes 
You're right about the regression, thanks for benchmarking it. I reproduced it and it's worse than
your 8MB case suggests: at a full 5MB chunk the ladder is ~10x slower than master, because plain
`append` grows by ~1.25x, so it regrows ~25 times and allocates ~28MB to hold 5MB.

I didn't take the 8KB→5MB jump, though. It removes the regrows, but it also gives up the win on
everything above 8KB: any larger response pays the same 5MB allocation as master (plus one copy),
so `query_range` on a normal Grafana panel — tens to hundreds of KB — gains nothing. And that band
is most of the reader's traffic.

The regrows are the real problem, not the small initial size. So the RAM chunk no longer grows a
single slice: it accumulates into a list of separately allocated blocks (8KB, then doubling up to
32KB — Go's largest size class). Blocks are never reallocated or copied, so the chunk allocates
only what is actually written, at any response size. They're reused after a flush to disk, and
the last block is clamped so the blocks tile the 5MB ceiling exactly.

Per request, end to end (`New` → writes of 4KB → full read → `Close`), same machine, one run:

| response | master | this PR (ladder) | 8KB→5MB jump | blocks (now) |
|---|---|---|---|---|
| 200 B | 565 µs · 5.01 MB | **1.3 µs · 8 KB** | **1.4 µs · 8 KB** | **1.3 µs · 8 KB** |
| 8 KB | 573 µs · 5.01 MB | **1.3 µs · 8 KB** | **1.4 µs · 8 KB** | **1.4 µs · 8 KB** |
| 64 KB | 661 µs · 5.01 MB | 25 µs · 208 KB | 625 µs · 5.02 MB | **12 µs · 88 KB** |
| 256 KB | 724 µs · 5.01 MB | 174 µs · 1.30 MB | 652 µs · 5.02 MB | **38 µs · 281 KB** |
| 1 MB | 566 µs · 5.01 MB | 1010 µs · 5.70 MB | 620 µs · 5.02 MB | **155 µs · 1.08 MB** |
| 3 MB | **535 µs · 5.01 MB** | 3680 µs · 18.15 MB | **569 µs · 5.02 MB** | **533 µs · 3.19 MB** |
| 5 MB | **563 µs · 5.01 MB** | 5561 µs · 28.60 MB | **625 µs · 5.02 MB** | 744 µs · 5.02 MB |

One honest regression is left: a response that fills the chunk exactly costs 744 µs against master's
563 µs (+32%), because 5MB in 32KB pieces has more per-allocation overhead than one 5MB block. It
doesn't grow past that — a response over 5MB spills to disk, where the file writes and the
ClickHouse query dominate. Below 2MB the win is 1.5x to 400x, and nothing anywhere is worse than
master except that one point.

Two variants I measured and dropped, so you don't have to re-check them. Doubling up to a threshold
and only then jumping to 5MB: the bytes the ladder allocated are wasted on the heavy path, 6.05MB for
a 5MB response at 719 µs, and the band just above the threshold is the worst of all (645 µs at 1MB).
Taking the whole remainder as one block once the response looks large: 947 µs at 5MB — one 4MB
allocation costs more than 120 small ones.

The benchmark is in the branch, so the last column is reproducible directly; the other columns need
the respective version of `ram_chunk.go`:

```
go test -run=NONE -bench=SmartBuffer -benchmem ./reader/utils/smart_buffer/
```

rita7lopes (changes requested):
I re-ran everything on my side and the sub-2MB band is exactly as good as you say. Confirmed the 5MB point too, on my machine it's a bit worse than your table (+59% instead of +32%, 963µs → 1535µs), but same shape, and I'm fine paying that.

What I did want to check is the part the benchmark doesn't cover: your `BenchmarkSmartBuffer` stops at exactly `chunkSize`, so nothing above 5MB is measured — and that's the spill-to-disk path, which is where the heavy `query_range` responses actually live. So I added 8MB and 20MB cases to the same harness and ran it against master and against your branch:

| response | master | this PR |
|---|---|---|
| 8 MB (spills) | 9.5 ms · 12 allocs | 15.5 ms · 175 allocs (**+63%**) |
| 20 MB (spills) | 21.0 ms · 12 allocs | 33.7 ms · 175 allocs (**+60%**) |

So the spill path got ~60% slower. I don't think it's the block design's fault though — it's `Flush` in `ram_chunk.go`: it does one `w.Write(block)` per block, and `fileBuffer.Write` is an unbuffered `os.File.Write`. That means every 5MB spill is now ~168 write syscalls where master did 1.

So before merging this one I think merging the writes in `Flush` is a good solution. Buffering inside `fileBuffer` is probably the cleanest. Also worth breaking out of the loop on the first empty block: after a `Clear` the block list is reused, so a small tail flush currently walks all ~160 leftover blocks and issues a zero-length write for each one.

When you do the benchmark again, remember to test sizes bigger than `chunkSize` to be able to really know if the code on the PR is a benefit comparing to main's approach to query_range queries

janeblower:
## What changed

- Blocks go back to a size-bucketed `sync.Pool` on `Close`, so the next response
  reuses warm memory instead of faulting in fresh pages.
- Block sizes are powers of two (so the pool can bucket them); the 5MB ceiling
  is enforced by byte count instead of tiling the blocks to it exactly.
- `Write` returns the total count: a write crossing the 5MB ceiling used to
  report only the tail, which reads as `ErrShortWrite` to `io.Copy`.

## Results

Real serve path: an HTTP server calling `SmartBufferServeStrings`, `GOMAXPROCS=4`,
separate load generator. Response mix: 78% 200B-2KB, 18% 32-512KB, 3.5% 1-4MB,
0.5% 6-12MB (spills to disk). Response length verified on every request.

Same offered load, 2000 rps (two runs each):

| | master | blocks | blocks + pool |
|---|---|---|---|
| server CPU | 65-70 s | 17.5-21.3 s | **11.6-13.6 s** |
| alloc per request | 4.9 MB | 174 KB | **4.8 KB** |
| GC cycles / total pause | 11840 / 2.3-2.6 s | 3070 / 0.44-0.56 s | **23 / 4-5 ms** |
| p99, 200B-2KB | 4.63 ms | 0.82 ms | **0.67 ms** |
| p99, 32-512KB | 4.87 ms | 1.53 ms | **0.90 ms** |
| p99, 1-4MB | 8.56 ms | 4.79 ms | **2.91 ms** |
| p99, 6-12MB (spills) | 22.8 ms | 14.9 ms | **11.9 ms** |
| peak RSS | 103-162 MB | 48-56 MB | 57-58 MB |

Saturated, 32 clients: master 3.8-4.2k rps, blocks 16.3-20.0k, blocks + pool 27.1-32.5k.

Against the blocks-only version: +65% throughput, -35% CPU, 36x less allocated,
130x fewer GC cycles, -40% p99 on heavy queries, -20% on the spill path. The
spill regression that needed buffered writes is gone: 11.9 ms against master's
22.8 ms.

Raising the block cap from 256KB to 1MB changes nothing outside the noise
(13.2 s vs 13.1 s CPU, 13.1 ms vs 13.7 ms p99 on spills), so the smaller cap
stays and the pool holds less memory per block.

The bench has no ClickHouse behind it, so the buffer dominates a request here.
In production the query time adds a constant on top and dilutes the latency win;
the CPU and garbage numbers do not depend on it.

rita7lopes (approved):
Approving. Re-ran everything on `a7a243b` against master `3622504` and the spill regression is gone. Dropping the full numbers here so we have a reference next time someone touches this file.

All runs: darwin/amd64, i7-8850H, go1.27.0, `benchstat`, n=6 (n=10 for the spill cases). Per request end to end — `New` → 4KB writes → full read → `Close`.

**In-RAM responses**

| response | master | this PR | |
|---|---|---|---|
| 200 B | 423.1 µs · 5.01 MB | **297.1 ns · 160 B** | −99.93% |
| 8 KB | 472.8 µs · 5.01 MB | **528.9 ns · 160 B** | −99.89% |
| 64 KB | 488.4 µs · 5.01 MB | **4.01 µs · 385 B** | −99.18% |
| 256 KB | 460.1 µs · 5.01 MB | **17.55 µs · 676 B** | −96.19% |
| 1 MB | 519.0 µs · 4.89 MiB | **81.13 µs · 1.13 KiB** | −84.37% |
| 3 MB | 725.5 µs · 4.89 MiB | **233.0 µs · 1.37 KiB** | −67.89% |
| 5 MB | 897.8 µs · 4.89 MiB | **436.7 µs · 2.49 KiB** | −51.36% |

All p=0.002. The exactly-5MB point that was +59% on the previous head is now 2x faster than master.

**Spill path (> 5MB, the one that regressed before)** — `-benchtime=60x`, n=10

| response | master | this PR | |
|---|---|---|---|
| 8 MB | 10.12 ms ± 6% | 11.73 ms ± 15% | +15.9% (p=0.002) |
| 20 MB | 26.51 ms ± 6% | 25.94 ms ± 3% | no difference (p=0.353) |
| B/op, both | ~4.89 MiB | **~11.4 KiB** | −99.77% |

Down from +63% / +60%. I checked whether the residual 16% at 8MB is still write fragmentation: patched `Flush` through a 1MB `bufio.Writer` on top of this head and got 11.73 → 11.17 ms (p=0.247) and no change at 20MB (p=0.481). So the 256KB block size already absorbed the syscall cost and buffering on top buys nothing — no further change needed here.

**Concurrent mixed traffic** — the closest thing to real reader load: `RunParallel`, sizes drawn per request, 60% ~400B label/series replies, 20% 20KB, 12% 200KB, 6% 1.2MB, 2% 4.8MB.

| | master | this PR |
|---|---|---|
| per request | 336.56 µs ± 3% | **17.52 µs ± 1%** (−94.79%, p=0.002) |
| allocated per request | 5,008,329 B | **370 B** (−99.99%) |

This is the number that decides it. Master allocates and faults 5MB per concurrent in-flight response regardless of its size — ten panels polling is 50MB touched for a few hundred KB of data. The pool holds up under contention with mixed size classes.

**One honest caveat: the 5MB win depends on a warm pool.** `sync.Pool` is drained on every GC, so I measured the first-request-after-GC case (two `runtime.GC()` per iteration, excluded from the timer):

| response | master cold | this PR cold | |
|---|---|---|---|
| 200 B | 500.93 µs | **12.80 µs** | −97.45% |
| 1 MB | 595.3 µs | **187.2 µs** | −68.55% |
| 5 MB | 935.5 µs | 1532.7 µs | **+63.84%** (p=0.002) |

So the old regression at exactly-5MB isn't eliminated, it's amortized away by the pool. It needs a response between ~4.7 and 5.0 MB landing on an empty pool, and it costs ~0.5 ms against a ClickHouse query of tens to hundreds of ms — fine to accept, just worth recording.

I checked whether `maxBlockSize` could tune it away, since that's the obvious lever (n=5, `-benchtime=300x`, cold):

| maxBlockSize | cold 5 MB | cold 1 MB | allocated for a 1 MB response |
|---|---|---|---|
| **256 KB (this PR)** | 1.44 ms | **188 µs** | **1.32 MB** |
| 512 KB | 1.33 ms | 236 µs | 1.59 MB |
| 1 MB | 1.40 ms | 332 µs | 2.11 MB |

It can't — all three sit at 1.3–1.4 ms against master's 0.94 ms, and larger blocks visibly hurt the 1MB band in both time and overshoot. The gap is inherent to ~24 separate spans vs one contiguous 5MB allocation, not to the constant. 256 KB is the right pick.

**Correctness checks I ran**, all under `-race`, all passing:
- 16 goroutines × 25 buffers, random sizes including 200B / 5MB / 5MB+1 / 7MB, randomly chunked writes, byte-exact readback and `Write` return-count assertions — no races, no corruption.
- Pooled blocks are not zeroed, so: write ~600KB of a marker payload, read, `Close`, then allocate a second buffer of the same size — no stale bytes come through. Only `block[:len]` is ever read or flushed, so the recycled capacity is unreachable.
- Read after `Close` returns `0, io.EOF` on the RAM path and a clean "file already closed" error on the spill path — `Release` nils the block slice, so a closed buffer can't reach memory it handed back to the pool.
- A single write straddling the 5MB ceiling returns the full count and the tail survives intact (`46a3a7f`).

`go vet` clean, `go build ./...` clean.

One nit, entirely optional: `blockClasses = 6 // 8KB..256KB` is only correct as long as `maxBlockSize` stays at 256KB — `blockClass` returns `-1` below 8KB and `6` at 512KB, both out of range for the array. Nothing can produce those sizes today since `nextBlockSize` is bounded by the same two constants, but I hit the panic myself by bumping `maxBlockSize` alone while running the sweep above. `const blockClasses = bits.Len(maxBlockSize / initialBlockSize)` would make the compiler enforce what the comment currently asserts.

The two `Read` details from my last review are still there and still non-blocking: an empty `p` returns `io.EOF` where master returned `0, nil` (no caller does that, `io.Copy` never does), and the last block returns `(n, nil)` then `(0, io.EOF)` on the next call instead of `(n, io.EOF)` — legal, costs one extra Read per response.

Nice work on this one — the block list plus the size-classed pool is the right shape, and the mixed-traffic number is a big win for the reader.

<details>
<summary>Benchmark harnesses used above (not in the branch)</summary>

```go
// Concurrent mixed traffic.
func BenchmarkMixedTraffic(b *testing.B) {
	mix := []int{}
	add := func(n, size int) {
		for i := 0; i < n; i++ {
			mix = append(mix, size)
		}
	}
	add(60, 400)      // labels / label values / series
	add(20, 20*1024)  // small query_range
	add(12, 200*1024) // normal Grafana panel
	add(6, 1200*1024) // wide panel
	add(2, 4800*1024) // near the 5MB ceiling
	data := make([]byte, 4096)

	b.ReportAllocs()
	b.RunParallel(func(pb *testing.PB) {
		rnd := rand.New(rand.NewSource(1))
		for pb.Next() {
			size := mix[rnd.Intn(len(mix))]
			buf := New()
			for written := 0; written < size; written += len(data) {
				if _, err := buf.Write(data[:min(len(data), size-written)]); err != nil {
					b.Fatal(err)
				}
			}
			if _, err := io.Copy(io.Discard, buf); err != nil {
				b.Fatal(err)
			}
			buf.Close()
		}
	})
}

// First request after a GC: two GCs clear sync.Pool including its victim cache.
func BenchmarkColdPool(b *testing.B) {
	sizes := []struct {
		name string
		size int
	}{{"200B", 200}, {"1MB", 1024 * 1024}, {"5MB", chunkSize}}
	data := make([]byte, 4096)

	for _, tt := range sizes {
		b.Run(tt.name, func(b *testing.B) {
			b.ReportAllocs()
			for b.Loop() {
				b.StopTimer()
				runtime.GC()
				runtime.GC()
				b.StartTimer()

				buf := New()
				for written := 0; written < tt.size; written += len(data) {
					if _, err := buf.Write(data[:min(len(data), tt.size-written)]); err != nil {
						b.Fatal(err)
					}
				}
				if _, err := io.Copy(io.Discard, buf); err != nil {
					b.Fatal(err)
				}
				buf.Close()
			}
		})
	}
}
```

The spill and in-RAM tables come from the branch's own `BenchmarkSmartBuffer` shape with the size list extended to `{200B, 8KB, 64KB, 256KB, 1MB, 3MB, 5MB, 8MB, 20MB}`.
</details>

PR #965, opened 2026-09-03
fix(logql): keep all lines of a multi-line stream on a single node

### LINKED ISSUE #964: LogQL single-node: multi-line log streams collapse to one line per stream
(h/t to my teammate @tanaykm for finding this issue)

## Summary

On a single node, a LogQL query that matches a stream with multiple log lines
returns only **one** line for that stream instead of all of them. The
regression was introduced by the label-resolution join change in
`fix(logql): drop log entries whose fingerprint has no labels yet` (commit
`2edfa85c71ee834fab50a42ee4f6d8f7ae13b3b6`, released in v5.4.3).

Both `/loki/api/v1/query` and `/loki/api/v1/query_range` are affected, because
they share the same label-resolution planner.

## Affected version

- v5.4.3 (and any build containing commit `2edfa85`).
- Earlier versions are not affected.

## Root cause

`LabelsJoinPlanner.Process`
(`reader/logql/logql_transpiler/clickhouse_planner/planner_labels_joiner.go`)
resolves stream labels by joining the samples subquery (`main`) to the
time-series/labels subquery (`_time_series`) on `fingerprint`.

The commit changed the single-node join from `ANY LEFT JOIN` to
`ANY INNER JOIN` to drop samples whose fingerprint has no labels row yet:

```sql
SELECT main.fingerprint, main.timestamp_ns, _time_series.labels,
       main.string, main.value
FROM main
ANY INNER JOIN _time_series
  ON main.fingerprint = _time_series.fingerprint
```

The relationship is one-to-many: `main` has **one row per log line** (many rows
per fingerprint), while `_time_series` has **one row per fingerprint**. In
ClickHouse, the `ANY` modifier keeps at most one matched pair per join key, so
`ANY INNER JOIN` collapses all the sample rows sharing a fingerprint down to a
single row. Every multi-line stream is reduced to one line.

The previous `ANY LEFT JOIN` preserved every left-hand (sample) row, so all
lines of a stream survived. Switching the join type conflated two different
goals — "drop rows whose fingerprint has no labels" and "keep only one row per
key" — and only the first was intended.

Note the cluster-mode branch was fixed differently, with an explicit
`WHERE length(labels) > 0` filter, and does **not** exhibit this collapse. Only
the single-node path regressed.

## End-to-end reproduction (single-node gigapipe)

1. Run a single-node gigapipe (`IsCluster = false`) backed by ClickHouse.
2. Push three log lines that share the same label set (one stream), e.g. via
   `POST /loki/api/v1/push` with three entries under a single stream such as
   `{k8s_node_name="test_node1"}`, plus one line under a second stream
   `{k8s_node_name="test_node2"}`.
3. Query them back within the query window:
   `GET /loki/api/v1/query?query={k8s_node_name=~"test_node[12]"}`
   (a range query over the same selector shows the same result).

**Expected:** the `test_node1` stream returns three entries; `test_node2`
returns one.

**Actual:** each stream returns exactly one entry. `test_node1` is missing two
of its three lines.

## Impact

Any single-node deployment silently drops log lines whenever a stream has more
than one line in the query window. This affects log browsing and any consumer
that reads raw stream entries, including downstream tooling that expects the
full set of lines for a selector.

## Suggested fix

Resolve labels without collapsing the many-side. Keep the samples-preserving
`ANY LEFT JOIN` and drop unresolved rows with an explicit filter, mirroring the
cluster branch:

```sql
FROM main
ANY LEFT JOIN _time_series
  ON main.fingerprint = _time_series.fingerprint
WHERE length(_time_series.labels) > 0
```

This still discards samples whose fingerprint has no labels row yet (the
original intent of the commit) while preserving every line of a resolved
stream. Verified against ClickHouse: the `WHERE length(...) > 0` filter drops
only the genuinely unresolved rows and returns all three lines of the
multi-line stream.

mgai-rchung:
## Summary

Fixes #964.

On a single node, a LogQL query that matches a stream with several log lines returns only **one** line for that stream instead of all of them. Both `/loki/api/v1/query` and `/loki/api/v1/query_range` are affected, since they share the same label-resolution planner.

## Root cause

`LabelsJoinPlanner.Process` resolves stream labels by joining the samples subquery (`main`) to the labels subquery (`_time_series`) on `fingerprint`. #961 changed the single-node join from `ANY LEFT JOIN` to `ANY INNER JOIN` to drop samples whose fingerprint has no labels row yet.

But the relationship is one-to-many: `main` has one row **per log line** (many per fingerprint), while `_time_series` has one row **per fingerprint**. ClickHouse's `ANY` modifier keeps at most one matched pair per join key, so `ANY INNER JOIN` collapses all sample rows sharing a fingerprint into a single row — every multi-line stream is reduced to one line.

The cluster-mode branch was not affected: it sheds unresolved rows with an explicit `WHERE length(labels) > 0` filter rather than by changing the join type.

## Fix

Keep the samples-preserving `ANY LEFT JOIN` on the single-node path and drop unresolved fingerprints with `length(_time_series.labels) > 0`, mirroring the cluster branch. This still discards samples whose labels are not yet visible (the intent of #961) without collapsing resolved multi-line streams.

```sql
FROM main
ANY LEFT JOIN _time_series
  ON main.fingerprint = _time_series.fingerprint
WHERE length(_time_series.labels) > 0
```

## Tests

- Updated the single-node case of `TestLabelsJoinDropsUnresolvedFingerprints` to assert the `length(_time_series.labels) > 0` filter instead of the `ANY INNER JOIN` keyword. (The prior string assertion is what let the regression through — it checked the join keyword, not the observable outcome.)
- Added `test/integration/logql_multiline_stream_test.go`: pushes a three-line stream and a one-line stream over `/loki/api/v1/push`, queries them back via `/loki/api/v1/query`, and asserts the line counts. It asserts behavior over HTTP, not SQL.

### Verification

Against a live gigapipe + ClickHouse stack built from this branch:

- **Buggy build** (`ANY INNER JOIN`): integration test fails — `multi-line stream returned 1 lines, want 3`.
- **Fixed build**: integration test passes; three-line stream returns 3, one-line stream returns 1.
- `go test ./...` passes; `gofmt` clean; static build succeeds.

The integration test runs in the existing `Integration` CI job (it provisions ClickHouse and runs `-tags integration ./test/integration/...`).

### DIFF
--- a/reader/logql/logql_transpiler/clickhouse_planner/planner_labels_joiner.go
+++ b/reader/logql/logql_transpiler/clickhouse_planner/planner_labels_joiner.go
@@ -107,6 +107,12 @@ func (l *LabelsJoinPlanner) Process(ctx *shared.PlannerContext) (sql.ISelect, er
 			AndWhere(sql.Gt(sql.NewRawObject("length(labels)"), sql.NewIntVal(0))), nil
 
 	}
+	// `main` holds one row per log line and joins many-to-one onto
+	// `_time_series` (one row per fingerprint). An ANY INNER JOIN keeps at most
+	// one matched pair per key and would collapse every multi-line stream to a
+	// single line, so the samples side is preserved with ANY LEFT JOIN and the
+	// unresolved fingerprints are dropped by the same length(labels) > 0 filter
+	// used above, rather than by the join type.
 	return sql.NewSelect().
 		With(withMain, withTS).
 		Select(
@@ -117,7 +123,8 @@ func (l *LabelsJoinPlanner) Process(ctx *shared.PlannerContext) (sql.ISelect, er
 			sql.NewSimpleCol("main.value", "value")).
 		From(sql.NewWithRef(withMain)).
 		Join(sql.NewJoin(
-			"ANY INNER ",
+			"ANY LEFT ",
 			sql.NewWithRef(withTS),
-			sql.Eq(sql.NewRawObject("main.fingerprint"), sql.NewRawObject("_time_series.fingerprint")))), nil
+			sql.Eq(sql.NewRawObject("main.fingerprint"), sql.NewRawObject("_time_series.fingerprint")))).
+		AndWhere(sql.Gt(sql.NewRawObject("length(_time_series.labels)"), sql.NewIntVal(0))), nil
 }
--- a/reader/logql/logql_transpiler/clickhouse_planner/planner_labels_joiner_test.go
+++ b/reader/logql/logql_transpiler/clickhouse_planner/planner_labels_joiner_test.go
@@ -88,7 +88,11 @@ func TestLabelsJoinBoundsSingleNodeJoinToMainFingerprints(t *testing.T) {
 }
 
 // A sample whose fingerprint has no time_series row yet belongs to no stream, so
-// it must not reach the client carrying an empty label set.
+// it must not reach the client carrying an empty label set. Both modes drop such
+// rows with a length(labels) > 0 filter. The single-node path keeps ANY LEFT
+// JOIN so it does not collapse multi-line streams (one row per line in `main`)
+// down to one row per fingerprint; the filter, not the join type, sheds the
+// unresolved rows.
 func TestLabelsJoinDropsUnresolvedFingerprints(t *testing.T) {
 	for _, tc := range []struct {
 		name    string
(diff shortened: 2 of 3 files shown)


mgai-rchung:
one question for the team: i added an integration test in the test/integration directory, but i see there are also many other e2e tests in a separate test repo. what is the recommended practice?

rita7lopes:
@mgai-rchung, thanks for the PR! The code is ok and the fix is right.
In terms of integration tests, we would prefer them to always live in the gigapipe-tests repo. Then we add a CI here to test both what exists now - a multi-node setup - and the new entry for the single-node test.

So, can you please delete the `test/integration/logql_multiline_stream_test.go`, or do you prefer me to remove it?

After that, this PR is good to go.

rita7lopes (changes requested):
`test/integration/logql_multiline_stream_test.go` should be removed

rita7lopes (approved):
We still will move the intergration test to the e2e-tests repo https://github.com/metrico/gigapipe-tests 
Until then, we will merge the PR as is to fix the main branch and create a fix release.

Thank you @mgai-rchung!

Issue #951, opened 2026-08-29
metrics_15s: keeps one row per raw metric sample and stores it 3.8× larger, and it cannot be turned off


janeblower:
## My case

We run gigapipe for metrics, and metric queries are served from `metrics_15s` — so 30 days of metrics means 30 days of the aggregate, and the only place left to cut is the source data. I hand-edit the TTL on `samples_v3` down to 1 day while `SAMPLES_DAYS` stays 30.

I dislike doing it: it throws away the source to keep a derived copy that holds nothing extra, `rotateTables` will silently overwrite it as soon as the configured policy changes, and it leaves one day of history for the paths that still read raw (`step < 15s`, unsupported functions).

## Problem

`metrics_15s_mv` pre-aggregates `samples_v3` into 15s buckets. That only helps when samples arrive faster than one per 15s — true for logs, false for a normal Prometheus scrape. At the conventional 15s interval every bucket holds exactly one sample, so the table aggregates nothing.

Measured on a stand (ClickHouse 25.3, Prometheus 3.1 at `scrape_interval: 15s`, 3 868 series, plus 240 000 log lines for comparison):

| signal | samples in `samples_v3` | rows in `metrics_15s` | ratio |
|---|---:|---:|---:|
| logs (`type=1`) | 240 000 | 2 410 | 99.6 : 1 |
| metrics (`type=2`) | 266 028 | 266 028 | **1 : 1** |

`SELECT count() FROM (SELECT fingerprint, timestamp_ns, countMerge(count) c FROM metrics_15s GROUP BY 1,2 HAVING c > 1)` returns `0` — not one bucket holds two samples, so `min`, `max` and `sum` are three more copies of the value already in `last`.

## Same data, 3.8× the space

Put both tables under the same sort order and codecs and a copy costs exactly what the original does: `samples_v3.value` is 0.748 B/row, `max`/`min`/`sum` are 0.748 B/row each, and `last` is 1.370 because an `argMaxState` carries the timestamp too — 4.36 B of payload for 0.748 B of data. Whole tables, same 266 028 samples: **`samples_v3` 1.01 B/row against `metrics_15s` 3.81 B/row.**

[...]

Issue #950, opened 2026-08-29
Log/trace payload columns ship without a compression codec, and the code meant to set one is dead


janeblower:
## Problem

`samples_v3.string` (log body), `time_series.labels`, `tempo_traces.payload`,
`profiles.tree`/`profiles.functions` and every column of `tempo_traces_attrs_gin`
are created with no `CODEC`, so ClickHouse falls back to LZ4.

There *is* machinery to make the codec configurable, but nothing can reach it:

1. **No env var.** All ClickHouse env vars are parsed in `portCHEnv`
   (`cmd/gigapipe/main.go:83-160`). `text_codec` / `logs_index` /
   `logs_index_granularity` are not among them, so the knobs exist only as
   config-file fields (`cloki-config/config/config.go:38-40`) and are not
   mentioned anywhere in `docs/`.
2. **No caller.** Those fields are read only by `RecodecDB`
   (`ctrl/qryn/maintenance/maintain.go:134`) and `ReindexDB` (`:143`), and
   neither function is called from anywhere — `ctrl/ctrl.go:16-20` wires up only
   `InitDB`, `UpgradeAll`, `RotateAll`.
3. **Wrong target table.** Even when invoked manually they fail:
   `ctrl/qryn/maintenance/codec.go:19,39,41` alter `samples_v4`, which the schema
   never creates (`ctrl/qryn/sql/log.sql:25` creates `samples_v3`).

Empirically, on a fresh instance `system.data_skipping_indices` is empty and
`SELECT * FROM settings WHERE type IN ('codec','index')` returns nothing — the
codec/index path has never run.

## Impact

Same data, wide parts, LZ4 (current) vs `ZSTD(1)`:

| column | LZ4 | ZSTD(1) | delta |
|---|---:|---:|---:|
| `samples_v3.string`, 50k log lines | 809.68 KiB | 484.94 KiB | −40.1% |
| `time_series.labels`, 50k series | 12.93 MiB | 7.45 MiB | −42.4% |
| `tempo_traces.payload`, 10k spans | 457.16 KiB | 377.06 KiB | −17.5% |
| `profiles.tree`, 60 profiles | 78.00 KiB | 41.01 KiB | −47.4% |

Whole tables on a mixed workload: `samples_v3` −16%, `tempo_traces` −23.5%,
`profiles` −29.8%. `ZSTD(3)` was measured too and is slightly *worse* than
[...]

PR #972, opened 2026-09-07
Split samples by signal

janeblower:
# Store metrics and logs in separate tables (opt-in)

Adds `SAMPLES_SPLIT_BY_SIGNAL`, which moves metrics out of the shared `samples_v3`
into their own table, and makes the metrics preaggregate tunable — on/off, bucket
width, retention. Off by default; with the flag unset nothing about the current
behaviour changes.

The motivation is that `metrics_15s` does not aggregate metrics. At a 15s bucket
and a 15s–30s scrape interval each bucket holds one point: measured on a real
cluster the view stores 41.6M rows for 42.6M samples, a 2% saving, while costing
as much disk as the raw data. It exists for logs, where it compresses ~100:1.
One `ORDER BY` and one MV serve both signals, so neither can be tuned for either.

## How it works

**Schema.** Four new tables in `log_split*.sql` (version keys 12/13/14), created
only when the flag is on: `samples_logs`, `samples_metrics`, `logs_aggr`,
`metrics_aggr`. `samples_metrics` has no `string` column and `metrics_aggr` has no
`bytes` — nothing on the metrics read path selects them. All carry codecs from the
start (`Delta`/`DoubleDelta`/`Gorilla` + `ZSTD(1)`).

**The metrics view is state-managed, not migrated.** Migrations here are
append-only with a per-file version counter: a statement runs exactly once, which
cannot express "disable" or "rebucket". `metrics_aggr_mv` is therefore driven from
a value stored in `settings`, the same idiom `rotateTables` already uses — compute
the desired state, compare, issue DDL only on a difference. A bucket-width change
must be a whole multiple of the previous one, or startup fails with an
instruction: stored rows are re-bucketed on read by `intDiv(timestamp_ns, step)`,
so they fold onto a coarser grid but never onto a finer one.

**Write path.** The row stream is split by `MType` in `IngestParsed`, the single
point every transport and protocol funnels through — splitting per handler would
be wrong, since influx and Loki-style push carry both signals in one request.
Reads filter `type IN (wanted, 0)`, so a dual-typed row (a Loki push with both a
line and a value) is written as a log row in one half and a metric row in the
other, and an undefined type stays 0 in both.

**Read path.** Table choice happens once, in `PopulateTableNames`, keyed on
`ctx.Type`, which every reader entry point already sets. Nothing below it knows
the split exists. The configured bucket width replaces the hardcoded 15s in the
step arithmetic, and with no aggregate present the PromQL optimizers are gated
off so accelerated queries fall back to raw samples instead of pointing at a view
that does not exist.

Also fixes `PushPromWriteRequest`, which duplicated the ingest loop inline and so
bypassed the splitter — under the split it would have written recording-rule
results where no reader looks, silently.

## Configuration

| variable | default | effect |
|---|---|---|
| `SAMPLES_SPLIT_BY_SIGNAL` | `false` | separate tables per signal |
| `METRICS_AGGR_ENABLED` | `true` | build `metrics_aggr`; when off, metric queries read raw samples |
| `METRICS_AGGR_INTERVAL` | `15s` | preaggregate bucket width |
| `METRICS_AGGR_DAYS` | `SAMPLES_DAYS` | preaggregate retention |
| `ADVANCED_METRICS_ORDERING` | `ADVANCED_SAMPLES_ORDERING` | `ORDER BY` for `samples_metrics` |

## Measurements

Stand: docker compose, ClickHouse 25.3 capped at 4 CPU / 4 GB / 200 MB/s disk.
Data: 42.6M raw samples, 54 432 series, 6h, replayed from a real
kube-prometheus-stack. Every layout below holds the same rows.

Disk, bytes per sample:

| | B/sample | vs master |
|---|---:|---:|
| master today (`samples_v3` + `metrics_15s`) | 22.57 | — |
| aggregate on, default order | 14.23 | −37% |
| aggregate off, default order | 9.82 | −56% |
| aggregate off, `ORDER BY (fingerprint, timestamp_ns)` | **1.19** | **−95%** |

Decomposed on identical rows: codecs alone −21.5%, sort order alone −83.3%, both
−90.5%. Codecs on the aggregate table −56%. Dropping the empty `string` column
saves nothing. Under the default order `fingerprint` is 64% of the table (7.98
B/sample); under `(fingerprint, timestamp_ns)` it is 0.02.

Query latency, wall ms / ClickHouse ms:

| query | aggregate on | raw, default order | raw, fp order |
|---|---|---|---|
| `avg_over_time[5m]`, 267 series | 147 / 79 | 1640 / 1555 | **125 / 38** |
| `sum(rate[5m])`, 2670 series | **439 / 429** | 770 / 552 | 699 / 481 |
| tail 5m, step 15 | 229 / 154 | **113 / 28** | 180 / 103 |
| instant, all series | 198 / 134 | **85 / 20** | 163 / 101 |

Two things follow. `ADVANCED_METRICS_ORDERING` is not optional once the aggregate
is off: without it a narrow selector scans the whole table, and that degrades
superlinearly — 54ms at 5.7M rows, 1555ms at 42.6M. With it, the same query goes
26ms → 38ms.

And the aggregate is not buying speed. Off + fp-order wins 4 of 7 query shapes,
loses on wide range aggregations by ~1.5x, and stores 12x less. Codecs are the one
real trade: +30% on a full scan for −21% disk if the default order is kept.

## Testing

`go test ./...` green with the flag off and with `SAMPLES_SPLIT_BY_SIGNAL=true`.
DDL, both materialized views and the ordering and aggregate switches were
exercised against a live ClickHouse.

## Limits

The flag is one-way in practice: once on, data already in `samples_v3` is not
read, and data written while on is not read after turning it off. Intended for new
installations; a manual backfill is possible with `INSERT ... SELECT` per day
partition. Re-enabling the aggregate does not backfill the window it was off for,
and queries over that window read as empty until the table is rebuilt.

### DIFF
--- a/cmd/gigapipe/main.go
+++ b/cmd/gigapipe/main.go
@@ -28,6 +28,7 @@ import (
 	rulerrouter "github.com/metrico/qryn/v5/ruler/router"
 	"github.com/metrico/qryn/v5/shared/commonroutes"
 	"github.com/metrico/qryn/v5/shared/distconfig"
+	"github.com/metrico/qryn/v5/shared/samplesconfig"
 	"github.com/metrico/qryn/v5/view"
 	"github.com/metrico/qryn/v5/writer"
 	writergrpc "github.com/metrico/qryn/v5/writer/grpc"
@@ -286,6 +287,9 @@ func main() {
 
 func start() {
 	distconfig.Init()
+	if err := samplesconfig.Init(); err != nil {
+		panic(err)
+	}
 	tables.InitDistTableNames()
 	var configPaths []string
 	if _, err := os.Stat(*appFlags.ConfigPath); err == nil {
--- a/ctrl/qryn/maintenance/aggr.go
+++ b/ctrl/qryn/maintenance/aggr.go
@@ -86,6 +86,24 @@ func aggrDropDays(configured, fallback int) int {
 	return configured
 }
 
+// parseAggrEnabled reads the enabled flag back out of a stored state string.
+// ok is false for an absent or unrecognised string, which the caller reads as
+// "no prior state" rather than "was disabled".
+func parseAggrEnabled(state string) (enabled bool, ok bool) {
+	for _, field := range strings.Fields(state) {
+		raw, found := strings.CutPrefix(field, "enabled=")
+		if !found {
+			continue
+		}
+		v, err := strconv.ParseBool(raw)
+		if err != nil {
+			return false, false
+		}
+		return v, true
+	}
+	return false, false
+}
+
 // SyncMetricsAggrMV brings metrics_aggr_mv in line with the configuration,
 // doing nothing when it already matches. It must run after the split tables
 // exist, since a materialized view needs both its source and its target.
@@ -114,6 +132,12 @@ func SyncMetricsAggrMV(db clickhouse.Conn, dbname, clusterName string,
 		return err
 	}
 	if samplesconfig.MetricsAggrEnabled() {
+		if wasEnabled, ok := parseAggrEnabled(have); !ok || !wasEnabled {
+			logger.Info(
+				"metrics_aggr_mv re-enabled: metrics_aggr holds nothing for the period " +
+					"it was absent (or has never run); queries over it will read that " +
+					"period as empty until it is rebuilt from samples_metrics")
+		}
 		logger.Info(fmt.Sprintf("Creating metrics_aggr_mv at %s buckets", interval))
 		if err := exec(metricsAggrMVQuery); err != nil {
 			return err
--- a/ctrl/qryn/maintenance/aggr.go
+++ b/ctrl/qryn/maintenance/aggr.go
@@ -1,0 +1,125 @@
+package maintenance
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/metrico/qryn/v5/ctrl/logger"
+	"github.com/metrico/qryn/v5/shared/samplesconfig"
+)
+
+// metricsAggrMVQuery is not part of the versioned migration on purpose. A
+// versioned script runs once, so it cannot express enabling, disabling or
+// re-intervalling the view; SyncMetricsAggrMV drives it from stored state
+// instead, the same way rotateTables drives TTL.
+const metricsAggrMVQuery = `CREATE MATERIALIZED VIEW IF NOT EXISTS {{.DB}}.metrics_aggr_mv {{.OnCluster}} TO {{.DB}}.metrics_aggr
+AS SELECT
+    fingerprint,
+    intDiv(samples.timestamp_ns, {{.AGGR_INTERVAL_NS}}) * {{.AGGR_INTERVAL_NS}} as timestamp_ns,
+    argMaxState(value, samples.timestamp_ns) as last,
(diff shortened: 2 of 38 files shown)


rita7lopes (changes requested):
Hello @janeblower,

This is interesting work and relevant, and it would be best used as part of smaller PRs were we can better measure each change and how it compares with master.

From your own issues #950 and #951 and also your own PR description saying: 
> "codecs alone −21.5%, sort order alone −83.3%, both −90.5%."

So the −95% is codecs plus sort order. 
Codecs are a metadata-only MODIFY COLUMN ... CODEC that re-encodes on merge. They work on the samples_v3 table in place, on every install that already exists. 
Sort order already ships: SAMPLES_ORDER_RUL defaults to "timestamp_ns" at ctrl/qryn/maintenance/update.go:212, samples_v3 is ORDER BY ({{.SAMPLES_ORDER_RUL}}) at ctrl/qryn/sql/log.sql:32, and ADVANCED_SAMPLES_ORDERING overrides it (cmd/gigapipe/main.go:123).

What I don't think is great about the PR:

1. It orphans logs. SamplesTable(false) returns samples_logs, so flipping the flag makes every log line in samples_v3 invisible on the next boot, for a change whose entire purpose is metrics. This is really bad for anyone using Gigapipe.
2. The split looks to be at the wrong place. "Which signal is this row" now lives in three layers: the parser decides the type (unmarshal.go:167 and :232, both flag-aware), the parser fans out time_series (unmarshal/builder.go), then the controller re-derives the split by rescanning MType and copying every row into two fresh TimeSamplesData (builder.go:297, six unpreallocated appends per row on a 1MB flush). The parser already knows the type when it builds the row. Having the controller scan it back out and memcpy the whole batch into two fresh TimeSamplesData (six unpreallocated appends per row, on a 1MB flush boundary) is redundant work at the wrong altitude. It's flag-gated so the default path pays nothing, but it makes the new code something that shouldnt exist.
3. The variable gate cannot be set to false after being used. The liability isn't the flag being on by accident. It's that once one user enables it, their data is readable only in that mode, so the flag can never be deleted and every future schema or read-path change has to be correct under both layouts, forever, with almost no test coverage on the second one.

What I would do instead is: go in small steps, smallest and widest-reaching first, measuring between each:
 - Codecs (#950). No flag, no new tables, lands on every existing install. Your own numbers: −21.5% on samples, −56% on the aggregate. Then we re-measure. This alone may be most of the answer. We measure the change against master.
 - Metrics out of the preaggregate (#951 Option 1). WHERE type != 2 via the append-only RENAME+recreate shape already used twice in log.sql, plus one boolean opt-out for sub-15s scrapes. −31% relative on top, nothing orphaned. We measure the change against master.
 - Sort order, as its own conversation. This is where "should we split by signal" actually belongs, because per-signal ordering is the only thing the split buys. If we do it, it needs a real migration (INSERT ... SELECT per day partition, dual-read window). This one is a big design change so let's coordinate before starting the code changes.

Let me know what you think! Happy to recheck if I got something wrong or missed somehing about the context and the end goal

janeblower:
Let’s leave this PR hanging for future reference as a PoC/rough draft for getting metrics memory usage down to ~1 byte/sample

rita7lopes:
Hello @janeblower,

This PR prompted a fair amount of internal discussion, and we're taking it as the foundation for improving the
metrics side of the repo, and logs and metrics will indeed be stored separately.

Thank you for all your contributions, especially this one

PR #975, opened 2026-09-09
Elastic: index labels

janeblower:
# Elastic ingest: `_index` and `_id` labels

## What this changes

The elastic controllers read `{target}` and `{id}` from a context key nothing in
the tree ever writes, so both were always empty. Every document ingested so far
carries `_index=""` and `_id=""`; a query like `{_index="app"}` never matched
anything. Because the labels were identical for every document, all elastic
ingest also shared a single fingerprint — one series for every index and every
id.

- `POST /{index}/_doc`, `POST|PUT /{index}/_create/{id}`, `PUT /{index}/_doc/{id}`
  now label the document with the index and, when the route has one, the id.
- A route without `{id}` no longer adds an empty `_id` label.
- `POST /{index}/_bulk`: the index in the action line wins over the one in the
  path, as in Elasticsearch. It used to be the other way round, invisibly — the
  path was always empty.
- An action line that resolves to no index at all — none in the path, none in
  the action — is now rejected with `400 index is missing`, as in
  Elasticsearch. `POST /_bulk` with `{"index":{"_id":"1"}}` used to be accepted
  and stored with no `_index` label.
- An action line carrying `_index` twice used to emit a duplicate key into the
  stored label set. The last value now wins and the key appears once.

Data written before this lands keeps the empty labels and lives in different
series, so anything filtering on `_index` sees a break at the upgrade. Ingest
that relied on a missing index being accepted starts failing instead.

## How this differs from Elasticsearch

Ingest only, into a log store — the endpoints accept the wire format, nothing
more:

- A document is stored as one log line: its raw JSON is the message, and the
  labels are `type=elastic`, `_index`, `_id` plus every string field of the bulk
  action line. Nothing is mapped, parsed, or indexed field by field.
- The timestamp is the time of ingest. `@timestamp` inside the document is
  treated as ordinary content.
- `_id` is a label, not an identity. Nothing deduplicates, overwrites, or
  versions: `_create` never conflicts, and re-sending the same id stores a
  second line. Ids that are unique per document also mean a series per
  document — keep them out of high-volume ingest.
- `delete` and `update` bulk actions are accepted and ignored.
- There are no per-item results, no per-document errors and no generated ids. A
  request that parses is answered `{"took":0,"errors":false}`; one that does not
  — a malformed action line, or an action with no index — fails as a whole with
  a 400. Elasticsearch also rejects a missing index for the whole request, but
  it rejects before storing anything: here, documents from earlier in the same
  request may already have been written. That is pre-existing behaviour for
  malformed input and is not addressed by this PR.
- A `type` key in an action line is dropped; it would collide with the
  `type=elastic` label.
- No search, mapping, template, or index-management API exists. Query the data
  through LogQL, e.g. `{type="elastic", _index="app"}`.

### DIFF
--- a/writer/controller/elastic.go
+++ b/writer/controller/elastic.go
@@ -37,30 +37,6 @@ func TargetDocV2(cfg MiddlewareConfig) func(w http.ResponseWriter, r *http.Reque
 			}))...)
 }
 
-//var (
-//	TargetDocV2 = Build(
-//		append(WithExtraMiddlewareDefault,
-//			withTSAndSampleService,
-//			withParserContext(func(w http.ResponseWriter, req *http.Request, parserCtx context.Context) (context.Context, error) {
-//				params := getRequestParams(req)
-//				// Access individual parameter values
-//				target := params["target"]
-//				id := params["id"]
-//				firstSlash := strings.Index(target, "/")
-//				if firstSlash != -1 {
-//					target = target[:firstSlash]
-//				}
-//				_ctx := context.WithValue(parserCtx, "target", target)
-//				_ctx = context.WithValue(_ctx, "id", id)
-//				return _ctx, nil
-//			}),
-//			withSimpleParser("*", Parser(unmarshal.ElasticDocUnmarshalV2)),
-//			withOkStatusAndJSONBody(200, map[string]interface{}{
-//				"took":   0,
-//				"errors": false,
-//			}))...)
-//)
-
 func TargetBulkV2(cfg MiddlewareConfig) func(w http.ResponseWriter, r *http.Request) {
 	return Build(append(cfg.ExtraMiddleware,
 		withTSAndSampleService,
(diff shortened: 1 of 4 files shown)


rita7lopes (approved):
Thanks for applying this fix @janeblower !!
I noticed while testing that Elastic rejects empty index names so therre's a one line to add only and that it is good to go

rita7lopes (changes requested):
Mistakenly approved on the other comment instead of asking for changes, but besides that line it's perfect!

rita7lopes (approved):
Verified this PR against `master`:

- **The tests pin the defect.** Checked out master's `writer/controller/elastic.go`, `writer/utils/unmarshal/elastic.go` and `writer/utils/context.go` into the branch, kept the new tests: all four fail. They pass on the branch.
- **Differential run, 30 request shapes** through the real router on both trees, comparing status, response body, series fingerprints, label sets and sample messages. Every difference is a fix or an intended change; nothing else moved.
- **Full suite**: 26 packages green on both, no new failures.
- **e2e**: `should write elastic` posts `{index:{_index:...}}` with no other action fields, so the label set is byte-identical before and after. The `should read elastic log` snapshot won't move.

The bug was worse than the description suggested: every `_doc` and `_create` document landed on the same fingerprint (`2594197576`) regardless of index or id.
One series for all elastic ingest, not just missing labels.

Two things the diff confirmed that are worth recording:

- Reordering `_index` to the end of the label slice does not split existing series. `fingerprintLabels` combines per-label hashes with `+`, `^` and `*`, all commutative, and the fingerprints come out identical on the cases that already had an action-line index.
- An action line with `_index` twice used to produce a duplicate JSON key in the stored label set. Fixed here as a side effect.

On rejecting a missing index with 400 rather than skipping the action: agreed, that's Elasticsearch's behaviour. It's a request-level validation failure there too, not a per-item one (elastic/elasticsearch#24319). Keeping the loud failure over a silent drop is the right call for an endpoint whose response carries no per-item detail.

One caveat I've added to the description rather than blocking on: a bulk that fails mid-stream can have written earlier documents, because the parser flushes at 1MB. Measured at 1025 of 2000 documents on a 1.9MB body. 
This predates the PR. Master does the same on a malformed action line, so it's a separate issue about bulk atomicity, not a regression here.

Thank you @janeblower !!

PR #976, opened 2026-09-14
Compress samples_v3 and metrics_15s columns

janeblower:
Split out of #972, which stays as is. That PR adds the split-by-signal layout and
is not ready; the codecs are independent of it and stand on their own.

`samples_v3` and `metrics_15s` carry the ClickHouse defaults on most columns:
`fingerprint`, `string` and every aggregate state are LZ4 with no specialised
codec, and `timestamp_ns`/`value` use `DoubleDelta`/`Gorilla` followed by LZ4.
This sets `ZSTD(1)` behind the numeric codecs, `Delta` on `fingerprint` and
`Gorilla` on the float aggregate states.

## Scope

Only the two `CREATE TABLE` statements change, and both already ran on every
existing installation — the migration runner replays `log.sql` from a stored
statement index, so an upgrade executes nothing new here. This is a default for
fresh databases; running deployments keep the layout they have and no data is
rewritten. Statement count is unchanged (31 before and after), so indices stay
aligned.

Distributed and `Merge` tables are untouched: they store nothing.

## Measurements

On a real kube-prometheus-stack replay (42.6M samples, 54 432 series, 6h,
ClickHouse 25.3 capped at 4 CPU / 4 GB / 200 MB/s disk) the codecs alone account
for −21.5% on the raw samples table and −56% on the aggregate.

Reproduced here on 10M synthetic rows (70% metric points, 30% log lines, same
rows inserted into both layouts, `OPTIMIZE FINAL` on each):

| table | master | this PR | |
|---|---:|---:|---:|
| `samples_v3` | 184.74 MiB | 162.11 MiB | −12.2% |
| `metrics_15s` | 302.91 MiB | 255.92 MiB | −15.5% |

Per column on `samples_v3`, bytes per row:

| column | master | this PR |
|---|---:|---:|
| `timestamp_ns` | 0.310 | 0.019 |
| `string` | 2.997 | 1.094 |
| `value` | 8.111 | 7.951 |
| `fingerprint` | 7.937 | 7.917 |

The synthetic numbers understate the real ones because the generated
fingerprints and float values are close to random; `fingerprint` in particular
only compresses once `ADVANCED_SAMPLES_ORDERING` sorts by it, and the mantissas
of real metrics repeat far more than generated noise does. `timestamp_ns` and
`string` are where the gain shows up regardless of the data.

The cost is CPU on decompression: ZSTD over LZ4 is roughly +30% on a full scan
of the raw table. Narrow selectors, which is what the read path actually issues,
are not measurably affected.

## Testing

`go build ./...` and `go test ./ctrl/... ./writer/... ./reader/...` green. Both
schema versions applied to a live ClickHouse 25.3 from an empty database, every
statement accepted; the materialized view fires and `metrics_15s` contents match
between the two layouts row for row and sum for sum.


rita7lopes:
Thank you for this, and for splitting it cleanly out of #972 — the analysis in both is genuinely excellent, and the issue you filed in #950 is what set this whole line of work in motion. The per-column byte measurements, the honest note about synthetic fingerprints understating the real numbers, and the decompression-cost caveat were all more rigorous than most schema PRs we see.

We are going to merge #977 for this one, and it's a scope decision rather than a quality one. Because this PR changes the two `CREATE TABLE` statements, the effect is limited to fresh databases: the migration runner replays from a stored statement index, so every existing deployment keeps LZ4 — which you note yourself in the description. #977 uses appended `MODIFY COLUMN` statements instead, so existing installs pick the codecs up on upgrade and converge on merge, and it covers traces, profiles, patterns and the gin tables in the same pass.

Where the two overlap, the choices measured out slightly differently on our side: on 2M rows, plain `ZSTD(1)` on `value` beat `Gorilla, ZSTD(1)` (11.7 MB vs 13.4 MB full-precision, 6.96 MB vs 13.0 MB rounded) under both the default ordering and `(fingerprint, timestamp_ns)` — `Gorilla`'s XOR output gives ZSTD less to work with. Your `Gorilla` on the `metrics_15s` aggregate states, on the other hand, is exactly what #977 landed on too.

So: closing this one as superseded, with real appreciation. The dead `codec.go` / `RecodecDB` / `ReindexDB` / `text_codec` cleanup you called for in #950 is next up as its own PR, and #972 stays open on its own merits — the split-by-signal measurements there deserve a separate discussion.

PR #977, opened 2026-09-14
fix(sql): add compression codecs to samples, traces, profiles and pat…

### LINKED ISSUE #950: Log/trace payload columns ship without a compression codec, and the code meant to set one is dead

44YHC:
(#950)

Append-only MODIFY COLUMN migrations; only new parts are affected, old parts stay readable and converge on merges.

Measured on 12h of realistic ingest (1.85M log lines, 2M metric samples, 650k spans, 455 pprof profiles):

- samples_v3               177.9 -> 111.6 MiB (-37%)
- tempo_traces_attrs_gin   191.5 -> 147.7 MiB (-23%)
- metrics_15s               79.2 -> 60.3 MiB (-24%)
- tempo_traces              60.0 -> 45.8 MiB (-24%)
- patterns                   0.9 -> 0.5 MiB (-48%)
- profiles.tree/functions  -33%/-38% (columns added by ALTER had no codec)

ZSTD(1) everywhere it wins; Delta+ZSTD for timestamps in tables not sorted by time; DoubleDelta+ZSTD where they are; span_id kept on LZ4 (random bytes, incompressible). Also evaluated ZSTD(3/6), LZ4HC, T64, GCD and FPC; none justified adoption (FPC -8% on value but raises the minimum ClickHouse to 22.8).


rita7lopes (approved):
Approving. Verified beyond CI, on a populated cluster rather than a fresh one, since e2e always boots an empty ClickHouse and so never exercises the upgrade path these statements actually take.

**Setup:** ClickHouse 25.3, 1 shard × 2 replicas with keeper, `Replicated*MergeTree` engines and real `ON CLUSTER` — the `replicated=true` branch in `ctrl/qryn/maintenance/update.go:233-236`. Master schema applied first (log/log_dist/traces/traces_dist/profiles/profiles_dist), then populated through the Distributed tables (300k samples, 50k series, 60k spans, 21k `metrics_15s` rows), then this branch's 11 statements applied as an upgrade.

**Migration mechanics**

- Statement counts 31→36, 8→11, 15→18; the runner's `";\n\n"` split (`update.go:110-120`) handles the appended comment blocks correctly, so existing version indices stay aligned.
- All 11 DDL entries `Finished` on both hosts with `exception_code 0`; 1.66s total, slowest statement 0.21s.
- `system.mutations` empty on both nodes, `system.replication_queue` empty, no replica went read-only, and `system.parts` byte counts were identical before and after (`samples_v3` 4,722,838 B both times). Metadata-only, as the description claims — confirmed on Replicated engines, not only plain MergeTree.
- `system.columns` identical across both replicas afterwards; row counts and checksums equal.
- Fresh-install path also applies clean (36/11/18 statements from an empty database).

**No schema regressions**

- `MODIFY COLUMN metadata String CODEC(ZSTD(1))` preserves `DEFAULT ''` on `time_series`, and the `type_v2` ALIAS columns are untouched.
- Every declared type matches the live column type exactly, including `tree Array(Tuple(UInt64, UInt64, UInt64, Array(Tuple(String, Int64, Int64))))` — a mismatch there would have triggered a full column rewrite; none occurred.
- `Gorilla` is accepted on the `SimpleAggregateFunction(max/min/sum, Float64)` columns (stored as `Gorilla(8)`).

**Effect confirmed**

- A part written after the ALTER: 300k rows in 818,215 B, against 4,722,838 B for the identical 300k rows written before it (−83%).
- Old parts converge on merge as stated: after `OPTIMIZE FINAL`, `tempo_traces` 503,010 → 295,711 B (−41%), `time_series` 463,601 → 236,006 B (−49%), `samples_v3` down to 1,636,408 B for 600k rows.

**Codec choices spot-checked** on 2M rows, both under the default `ORDER BY timestamp_ns` and under `(fingerprint, timestamp_ns)`. Dropping `Gorilla` from `value` in favour of plain `ZSTD(1)` holds up in both (11.7 MB vs 13.4 MB with `Gorilla, ZSTD(1)` on full-precision floats; 6.96 MB vs 13.0 MB on rounded ones) — `Gorilla`'s XOR output feeds ZSTD poorly. Plain `ZSTD(1)` on `fingerprint` is likewise right given the default ordering is by timestamp.

**Costs measured**

- Ingest: no measurable penalty (2M rows, master 0.86–0.89s vs this branch 0.67–0.94s).
- Reads: decompression-bound full scans get roughly 1.7x slower — `WHERE string LIKE '%…%'` over 2M rows went 0.11–0.13s to 0.19–0.22s. Expected for ZSTD over LZ4 and worth it at these ratios; could you add a line about it to the description so it's on the record?

**One operational note, not a blocker:** with a replica down, an `ALTER ... ON CLUSTER` returns `Code: 159 TIMEOUT_EXCEEDED` after `distributed_ddl_task_timeout` (180s by default), which propagates to `panic(err)` at `cmd/gigapipe/main.go:74-77` and fails the boot. It is fully recoverable — `ver` is only written after a successful exec so the statement is retried, the offline replica applies the queued DDL when it returns, and re-running the same ALTER is idempotent (verified). Every pre-existing migration statement in these files behaves the same way; these 11 just widen the window a little. Worth knowing during rolling upgrades.

Optional follow-up: `time_series.name` is still on LZ4 and would compress well. The dead `codec.go` / `RecodecDB` / `ReindexDB` / `text_codec` cleanup that #950 also asks for is coming as its own PR.

PR #979, opened 2026-09-14
fix(reader): floor query_range end to the 15s grid instead of ceiling it

bzed:
## Problem

`reader/controller/prom_query_range.go` rounded the `query_range` `end` parameter **up** to the next multiple of 15s (`math.Ceil`) before handing it to the PromQL engine as the literal `End` of the range query:

```go
req.End = time.Unix(int64(math.Ceil(float64(req.End.Unix())/15)*15), 0)
```

Since the engine evaluates a data point at every `Start + k*Step <= End`, rounding `End` *forward* fabricates one extra timestamp strictly after what the caller asked for, whenever `end % 15 != 0`.

### Verification

Pushed identical synthetic data (a steady counter, 1/s, sampled every 15s) into both gigapipe and a real Prometheus 3.14 instance (via podman, same `remote_write` payload), then compared `query_range` results for `rate`/`irate`/`deriv`/`delta`/`idelta`/`increase`/`resets` at many steps. Every function showed the same extra timestamp in gigapipe's output (`end + (15 - end%15)`) whenever `end % 15 != 0`; real Prometheus never returns a point after the requested `end`. This reproduced identically across both PromQL code paths (ClickHouse-pushdown `CounterPlanner` and the older `DownsampleHintsPlanner`), confirming it's a controller-level rounding bug rather than something specific to one function.

## Fix

`Start` was already floored to the 15s grid (`req.Start.Unix()/15*15`) — `End` should follow the same direction, not the opposite one. Flooring both bounds:
- keeps `Start <= End` (floor is monotonic)
- preserves alignment to the `metrics_15s` table's native resolution
- guarantees the response window never extends past what was requested, matching real Prometheus

Extracted the rounding into a small `snapQueryRangeToNativeResolution` helper (in [reader/controller/prom_query_range.go](reader/controller/prom_query_range.go)) with a comment documenting the invariant, and switched `End` from `math.Ceil` to floor.

The original line traces back to a squashed `qryn -> gigapipe` merge commit with no isolated rationale preserved in this repo's history; the fix follows `Start`'s existing (correct) direction instead.

## Tests

Added `reader/controller/prom_query_range_test.go`:
- an invariant sweep over 30 sub-15s offsets asserting the snapped `end` is never after the requested `end` (and both bounds land on the 15s grid, with `start <= end` preserved)
- an "already 15s-aligned bounds pass through unchanged" case
- the exact reported scenario (`end % 15 == 7`) asserting it floors to the prior boundary instead of the old ceiled one

`go build ./...` and `go test ./reader/controller/... ./reader/service/... ./reader/promql/...` all pass.


rita7lopes (approved):
Hey @bzed, 

Thanks for the PR! 
Indeed the engine emits a point at every Start + k*Step <= End, so ceiling End was the only variant that could put a timestamp past the requested window. Flooring it produces the same series as not snapping End at all, which matches Prometheus. 

Merging

PR #981, opened 2026-09-14
fix(promql): rate/irate/increase/delta/deriv/resets return no series when step >= range

### LINKED ISSUE #980: query_range (and [range:resolution] subqueries) return zero series for range-vector functions when step >= range
## Summary

On a gigapipe deployment updated on 2026-09-14, `/api/v1/query_range` (and
the PromQL `[range:resolution]` subquery form used inside an instant
`/api/v1/query`) returns an **empty result** (`{"status":"success","data":
{"resultType":"matrix","result":[]}}`) for **any** range-vector function
(`rate()`, `irate()`, `increase()`, `delta()`, `deriv()` all confirmed)
whenever the query's own step (the HTTP `step` parameter, or the
subquery's own resolution) is **greater than or equal to** the function's
range-vector duration — e.g. `rate(x[300s])` queried with `step=300s`.

This reproduces identically on two independent, unrelated gigapipe
deployments (both updated within the last few days), and did **not**
happen before the update on at least one of them (regression, not a
pre-existing limitation we simply hadn't hit before).

The same expression evaluated with **no** step at all (a plain instant
`/api/v1/query`) returns correct, complete data every time. A bare
(non-function) selector via `query_range` at the same step is also
unaffected — only a range-vector function's own result is empty.

## Impact

Any client (ours included — see below) that builds `step_seconds ==
rate_window_seconds` for a `query_range`/subquery call — a completely
standard, idiomatic PromQL pattern (exact back-to-back tiling of
successive rate windows, no gaps or overlap) — gets zero data back,
silently: HTTP 200, `"status":"success"`, just an empty `result` array. No
error, no warning field in the response. This is a very easy trap: the
underlying counter data is present and healthy (confirmed directly, see
below) the whole time; only the derived range-vector-function query is
affected.

## Reproduction

Three `curl` calls against the same live endpoint, same metric, same time
range, differing only in `step`:

```sh
BASE="http://<gigapipe-host>:<port>"
NOW=$(date +%s)
START=$((NOW-3600))

# step < range: works
curl -sS --get "$BASE/api/v1/query_range" \
  --data-urlencode "query=rate(some_counter[300s])" \
  --data-urlencode "start=$START" --data-urlencode "end=$NOW" \
  --data-urlencode "step=299s"
# -> data.result has N series, each with real values

# step == range: broken
curl -sS --get "$BASE/api/v1/query_range" \
  --data-urlencode "query=rate(some_counter[300s])" \
  --data-urlencode "start=$START" --data-urlencode "end=$NOW" \
  --data-urlencode "step=300s"
# -> data.result: []

# step > range: also broken
curl -sS --get "$BASE/api/v1/query_range" \
  --data-urlencode "query=rate(some_counter[300s])" \
  --data-urlencode "start=$START" --data-urlencode "end=$NOW" \
  --data-urlencode "step=301s"
# -> data.result: []
```

### Exact boundary, binary-searched

Holding the range-vector duration fixed at `[300s]` and varying only
`step`:

| `step` | series returned |
|---|---|
| 60s – 299s | full data (72 series in our test) |
| **300s** | **0** |
| 301s – 400s (tested up to) | **0** |

Independent of absolute step size and of window length — a 20-minute
window, a 1-hour window, and a 24-hour window all show the identical
pattern at the identical boundary.

### It's the step/range *relationship*, not either value alone

Holding `step=300s` fixed and varying the range-vector duration instead:

| query | series returned |
|---|---|
| `rate(x[600s])` at `step=300s` (step < range) | full data |
| `rate(x[300s])` at `step=300s` (step == range) | **0** |
| `rate(x[299s])` at `step=300s` (step > range) | **0** |
| `rate(x[301s])` at `step=300s` (step < range, by 1s) | full data |

So the condition is precisely `step >= range`, not a fixed threshold on
either number.

### Instant queries are unaffected

```sh
curl -sS --get "$BASE/api/v1/query" \
  --data-urlencode "query=sum by (some_label) (rate(some_counter[300s]))"
# -> full, correct data, every time -- no "step" parameter exists on this endpoint
```

### The PromQL subquery form (`[range:resolution]`) has the identical bug

This is evaluated via `/api/v1/query` (an instant query), with the
resolution embedded in the query text itself rather than as an HTTP
`step` parameter — and it fails the same way:

```sh
# resolution (300s) >= the inner rate()'s own range (300s): broken
curl -sS --get "$BASE/api/v1/query" \
  --data-urlencode "query=quantile_over_time(0.95, (rate(some_counter[300s]))[1h:300s])"
# -> data.result: []

# resolution (299s) < range: works
curl -sS --get "$BASE/api/v1/query" \
  --data-urlencode "query=quantile_over_time(0.95, (rate(some_counter[300s]))[1h:299s])"
# -> full, correct data
```

### The underlying counter data is not the problem

A **bare** range query (no `rate()`/function wrapper at all) at the same
step, same window, returns complete, continuous, correctly-typed numeric
data:

```sh
curl -sS --get "$BASE/api/v1/query_range" \
  --data-urlencode "query=some_counter" \
  --data-urlencode "start=$START" --data-urlencode "end=$NOW" \
  --data-urlencode "step=300s"
# -> full data, every series, real numeric values at steady scrape-interval spacing
```

This rules out "not enough history yet" / a data-availability gap as the
cause — the raw samples the `rate()` window would need are demonstrably
present and correctly typed the whole time; only the range-vector
function's own evaluation over `query_range`/the subquery form is
affected.

### Functions tested, all reproduce identically

`rate()`, `irate()`, `increase()`, `delta()`, `deriv()` — same boundary,
same symptom, for every one of them.

### `sum by (...)` alone (no range-vector function) is unaffected

```sh
curl -sS --get "$BASE/api/v1/query_range" \
  --data-urlencode "query=sum by (some_label) (some_counter)" \
  --data-urlencode "start=$START" --data-urlencode "end=$NOW" \
  --data-urlencode "step=300s"
# -> full data
```

So the bug is specific to a *range-vector function's* own evaluation
inside `query_range`/the subquery form, not to aggregation or to
`query_range` in general.

## Environment

- Reproduced independently on two separate gigapipe deployments, each
  fed by Telegraf → InfluxDB line protocol → gigapipe's
  Prometheus-HTTP-API-compatible read endpoint.
- Both deployments were **recently updated** (within the days before this
  was found); on at least one of them, the identical query shape
  (`step == rate_window`, i.e. the client's normal, unmodified behavior)
  worked correctly before that update — this looks like a **regression**,
  not a preexisting limitation.
- `/api/v1/status/buildinfo` on both deployments returns
  `{"status":"success","data":{"version":""}}` — an **empty** version
  string, so we cannot cite an exact gigapipe build/version number from
  the API itself. Whoever manages the deployment should be able to pull
  the actual version from their own deployment records/image tag.
- Metric shape: Telegraf's InfluxDB output, so labels include `vmid`,
  `instance`, `nodename`, `host`, `measurement`, `object`, `service_name`
  in addition to the metric name — unremarkable, standard
  Telegraf-via-InfluxDB-line-protocol tagging, not believed to be
  relevant to the bug (the same symptom reproduces on a trivially simple
  synthetic query shape too).

## Why this is easy to hit in practice

`step == rate_window` (exact, back-to-back tiling — the range-vector
window for one evaluated point ends exactly where the next one's begins,
covering the whole queried span with no gaps and no overlap) is a
completely standard PromQL pattern, not an edge case. Any client whose
config happens to set these two independently-named knobs to the same
value — which is a very natural default (we did, ourselves, before
finding this) — silently gets no data back from any `query_range`/subquery
call it makes, with no error to signal why.

## Suggested workaround (client-side, until fixed)

Until this is fixed: any `step`/subquery-resolution used with a
range-vector function must be kept **strictly less than** that function's
own range-vector duration — e.g. request a step that is an exact integer
divisor of the intended step and strictly below the rate window, and (if
the caller needs the coarser grid back) decimate the denser response down
by keeping every Nth point afterward. We've applied exactly this as a
defensive workaround in our own client (`proxmox-storage-drs`) so it keeps
working against this deployment shape regardless of when/whether this is
fixed upstream.

bzed:
Fixes #980

## Summary

`rate()`, `irate()`, `increase()`, `delta()`, `deriv()` (and, by
construction, `resets()`/`changes()`/`idelta()`) returned an **empty**
`query_range` result — HTTP 200, `"status":"success"`, no error — for
any series whenever the query's `step` reached or exceeded the
function's own range-vector duration, e.g. `rate(x[300s])` queried
with `step=300s`. That's an ordinary, standard PromQL shape (`step ==
range`, back-to-back tiling), not an edge case, so any client whose
step happens to match its rate window silently got nothing back. See
#980 for the full reproduction and a binary-searched boundary.

## Root cause (two independent code paths, same underlying flaw)

1. **The ClickHouse pushdown** (`VectorRange` optimizer →
   `CounterPlanner`/`CounterFlagsPlanner`, used for `rate`, `increase`,
   `delta`, `resets`, `changes`): real samples were bucketed once per
   `ctx.Step` before being scanned for two distinct samples inside
   `(t-range, t]`. Once `step >= range`, every bucket landed exactly on
   the step grid, so that window could hold at most one of them —
   "two distinct samples" was never satisfied, and the result was
   always empty.

2. **The older per-step resampling path** (`DownsampleHintsPlanner`,
   used for `irate`, `deriv`, `idelta`, which have no ClickHouse-side
   acceleration and rely on the Prometheus engine's own function
   implementation over whatever the storage layer hands back): same
   idea, but worse — the resample bucket grid is anchored to epoch,
   not to the query's own evaluation timestamps, so whether a given
   `(t-range, t]` window catches enough buckets to compute anything
   depends on an alignment the caller never controls. Verified live
   against ClickHouse: `irate(x[300s])` failed unpredictably well
   before `step >= range` too (empty at step 210s, 240s, 260s... while
   working at 220s, 230s, 250s in the same sweep), which is why the
   `step >= range` boundary in the report is necessary but not
   sufficient to describe it.

## Fix

Both paths get the same fix: cap the resample/bucket width to at most
`range/2` for functions that measure a change across samples (`rate`,
`irate`, `deriv`, `delta`, `idelta`, `resets`; `increase`/`changes`
included defensively). Two buckets of that width always fit inside any
`(t-range, t]` window, regardless of where its boundary falls — a
pigeonhole guarantee, not something that depends on `step` or on
alignment. `ctx.Step` is used unchanged whenever it already leaves
room for two buckets, so the common case (step finer than range) is
untouched.

`OverTimePlanner` (`sum_over_time`, `avg_over_time`, `min_over_time`,
`max_over_time`, `count_over_time`, `last_over_time`) deliberately does
**not** get this cap: those functions only need one sample in the
window, not two, so there's no correctness floor forcing anything
finer than `ctx.Step`, and doing so anyway would raise their query cost
for no reported benefit. See the last commit and
`TestOverTimeKeepsQueryStepEvenWhenCoarserThanRange`.

## Performance

The cap only changes behavior once `step >= range/2` for the affected
functions, and even then the cost is bounded by, and identical to, an
already-supported query shape: a query with `step = range/2` explicitly
(never broken, already accepted). A caller with `step = 1000×range`
does not pay 1000× anything — they pay exactly what a `step = range/2`
caller already pays. No new class of expensive query is introduced;
correctness for a previously-silently-empty shape now costs the same
as an already-working neighboring one.

## Verification

- Unit tests added per fix (fail against the pre-fix code, pass with
  it): `TestCounterAcceleratesWhenStepIsNotFinerThanRange`,
  `TestDownsampleHintsCapsChangeFunctionBucket` /
  `TestDownsampleHintsLeavesPlainAggregatesAtTheQueryStep`,
  `TestOverTimeKeepsQueryStepEvenWhenCoarserThanRange`.
- Verified end-to-end against a real, local ClickHouse: pushed a
  synthetic counter via remote_write, swept `step` from 30s–500s across
  `rate`/`irate`/`deriv`/`delta`/`idelta`/`increase`/`resets`, confirmed
  every function returns data with correct values at every step
  (previously empty at/around `step >= range`).
- Cross-checked against a real Prometheus 3.14 fed the same data:
  `rate`/`delta`/`increase`/`resets` match Prometheus byte-for-byte
  across a 75-step sweep. `irate`/`deriv` match closely but not always
  exactly, and `idelta` numerically differs at most resolutions — both
  are a **pre-existing, separate** limitation of
  `DownsampleHintsPlanner` (it hands the engine a resampled proxy
  series tagged with the bucket boundary rather than the real sample
  timestamp, which these two-point functions are sensitive to),
  present identically whether or not this fix's cap is even engaged,
  so out of scope for this PR.


rita7lopes (approved):
Thanks for the PR @bzed !
The test `TestOverTimeKeepsQueryStepEvenWhenCoarserThanRange` asserts a current behaviour that is wrong. 
sum_over_time(x[1m]) at step=1h buckets at 1 hour, so it sums a whole hour of samples and calls it a 1-minute sum. Basically it gives wrong numbers and the test asserts it. 

This bug is in master so I will not hold this PR which solves a real (different) bug because of something it doesn't introduce. 
I will fix the sum bug in a coming PR.

Merging this one

PR #998, opened 2026-09-19
fix(promql): key selector buckets by the ceiling, like every other read

rita7lopes:
> **Stacks on #984.** Based on `fix/promql-over-time-window`; GitHub will retarget it to `master` once that merges.

#984 corrected the bucket keying a `BucketProducer` renders, so a bucket holds `(key-width, key]` and buckets tile backwards from the timestamp being evaluated. The **other** read of `metrics_15s` was left keying by the floor.

## The change

A query carrying no range function is planned by `DownsampleHintsPlanner`'s `default` branch, not by a `BucketProducer`, and that branch keyed by the floor. It now uses the same keying, through one definition — `bucketTimestampCol` — applied at both sites.

The width stays the caller's: a range function sizes its bucket against its range (`BucketResolution`), a bare selector against the query step. The rule that *a bucket ends at its key* is one decision, and duplicating it is what let #984 correct one path and leave this one behind.

## Measured

Two builds of this branch — with and without the fix — reading the **same** ClickHouse, against a reference Prometheus 2.55.1 fed by `remote_write`. `demo_memory_usage_bytes{type="buffers"}`, compared on shared timestamps.

Count of points whose value equals Prometheus's value one step later (a one-step shift):

| step | without fix | with fix |
|---|---|---|
| 15s | 20 of 20 shifted | 20 of 20 shifted |
| 30s | 16 of 16 shifted | **0 shifted** |
| 60s | 8 of 8 shifted | **0 shifted** |
| 120s | 4 of 4 shifted | **0 shifted** |

The reader-side regrouping shift is removed for every step coarser than the bucket.

**At step == 15s the change is a no-op, by construction.** `metrics_15s` rows are already stamped on a 15s grid, so for a row whose timestamp is an exact multiple of 15s, `intDiv(ts + 15s-1, 15s) * 15s` and `intDiv(ts, 15s) * 15s` are the same key. The table shows exactly that.

## What this does NOT fix

Values still do not match Prometheus at any step, and this PR does not claim otherwise. The residual error is one *bucket*, and it originates at the write side, in `ctrl/qryn/maintenance/metrics15s.go:20`:

```sql
intDiv(samples.timestamp_ns, 15000000000) * 15000000000 as timestamp_ns,
argMaxState(value, samples.timestamp_ns) as last,
```

The materialized view stamps each row by the floor, so a row labelled `T` holds samples from `[T, T+15s)` and its `last` is the newest sample in that interval — up to 15s after `T`. At a 60s step, Prometheus returns the sample at `t` while gigapipe returns the sample at `t+15s`, read out of the row labelled `t`.

No reader-side keying can correct rows that are already mislabelled, so that belongs to its own change — one that has to reckon with rebuilding the aggregate. Note #972 carries the same stamping forward, parameterized as `AGGR_INTERVAL_NS`.

## Scope

The `default` branch also serves the reducers the engine evaluates itself (`sum_over_time` and friends) when they are not accelerated; they are regrouped by the same expression and move with it.

`TestDownsampleHintsLeavesPlainAggregatesAtTheQueryStep` asserted the rendered floor expression while meaning to assert that the bucket width is the uncapped step. It now asserts that through the same helper, so it survives a change of keying rather than pinning one.

## Verification

- `go build ./...` clean, `gofmt` clean
- `go test ./reader/... -count=1` — 0 failures
- `TestSelectorBucketCoversTheIntervalEndingAtItsKey` fails with the fix reverted, passes with it
- The step table above, measured against a real Prometheus

## Not included

`DownsampleValuesPlanner` patches `timestamp_ms` and then `DownsampleHintsPlanner` patches the same alias over it, so the first is dead for this path. Left alone rather than changed silently.


44YHC:
Reviewed the whole stack (#983 → #984 → this). Direction is right and the three read as one arc: one definition of the bucket *width* (#983), one definition of what a bucket *covers* (#984), applied to the read that got missed (here). Verified at `ae87d1e`: `go build ./...` clean, `go test ./reader/... -count=1` exit 0, 16 packages, no failures.

One thing to pick up before this merges.

## The third call site is still floor-keyed

`bucketTimestampCol` is introduced as "one rule … so it cannot be corrected in one read path and left behind in another", then applied to two of the three sites in `downsample_hints.go`. The `NeedsDistinctSamples` branch still renders the floor form by hand:

https://github.com/metrico/gigapipe/blob/ae87d1eca4547b291059ea15913f2411371b6376/reader/promql/promql_transpiler/planner/downsample_hints.go#L62-L63

Rendered side by side off this branch's tip:

```
deriv  step=60s range=300s:  intDiv(samples.timestamp_ns, 60000 * 1000000) * 60000
bare selector step=60s:      intDiv(samples.timestamp_ns + 59999999999, 60000000000) * 60000
```

Same planner, same table, same `argMaxMerge(samples.last)` — and this PR's own argument applies verbatim: the bucket holds `[key, key+step)`, `last` is the newest sample in it, so the value reported at `t` is the one belonging to `t+step`.

The branch is live. `deriv` and `changes` are absent from `supportedFunctions`, so `ok == false`, `useRawData` stays false, no substitute is registered, and they land here — exactly the path #983's description identifies as the one `deriv` takes. The fix looks like a one-line swap to the same helper the `default` branch uses:

```go
timeField := bucketTimestampCol("samples.timestamp_ns",
    time.Duration(step)*time.Millisecond) + compat4019
```

(`step` there is already a `time.Duration` before `.Milliseconds()`, so it's a small rearrangement.)

## Two things that corroborate it was an oversight rather than a decision

- `TestDownsampleHintsCapsChangeFunctionBucket`'s `uncappedBucket` helper pins the literal floor expression for this branch — the precise anti-pattern this PR calls out and fixes in `TestDownsampleHintsLeavesPlainAggregatesAtTheQueryStep`, applied to the sibling test but not this one.
- #984's `TestBucketCoversTheIntervalEndingAtItsKey` covers `rate` / `increase` / `resets` only through the `BucketProducer` path, so nothing currently guards the `DownsampleHintsPlanner` change-function branch. A case there (`deriv`, which has no pushdown) would have caught this and would keep catching it.

## Minor, on the scoped-out item

Agree with leaving `DownsampleValuesPlanner` alone, and the "it's dead for this path" reasoning is right — but it's only dead while `Hints.Step != 0`. The early return at `downsample_hints.go:24` leaves both its floor keying *and* an `intDiv(x, 0)` standing if a zero step ever reaches it. `adjustHintsForRate` prevents that today; worth a line in whatever picks that up later so the guarantee is written down somewhere other than a different file.

---

Everything else reads well. The step table measured against a real Prometheus is the right kind of evidence, and calling out explicitly that step == 15s is a no-op by construction — rather than quietly showing a row that looks like a failure — is the detail that makes the table trustworthy. The write-side `metrics15s.go` stamping being named as the residual, with the note that #972 carries it forward, is useful context to have recorded here.

44YHC (approved):

